PR #3, opened 2019-05-09
Improved sanitization and formatting.

### LINKED ISSUE #2: sanitize html: only top-level nodes are sanitized
e.g. `'<div><div id="x"></div></div>'` is returned back as the same string as sanitized html

isoos:
- fixes #2 
- no longer depends on `universal_html`:
- custom output formatting that keeps no-content tags as self-closing: ``

I've wanted to create a separate library for the formatting (e.g. allow indents and a canonical formatting of the html markup), but the edges cases are hard to deal with, and I've settled with this simple solution.

The runZoned part could be probably removed? /cc @jonasfj

### DIFF
--- a/sanitize_html/lib/sanitize_html.dart
+++ b/sanitize_html/lib/sanitize_html.dart
@@ -14,15 +14,7 @@
 
 library sanitize_html;
 
-import 'dart:async' show Zone, ZoneDelegate, runZoned, ZoneSpecification;
 import 'src/sane_html_validator.dart' show SaneHtmlValidator;
-
-void _printHandler(Zone self, ZoneDelegate parent, Zone zone, String line) {
-  // Suppress printed lines about stuff being removed.
-  if (!line.startsWith('Removing disallowed')) {
-    parent.print(zone, line);
-  }
-}
 
 /// Sanitize [htmlString] to prevent XSS exploits and limit interference with
 /// other markup on the page.
@@ -60,10 +52,8 @@ String sanitizeHtml(
   bool Function(String) allowElementId,
   bool Function(String) allowClassName,
 }) {
-  return runZoned(() {
-    return SaneHtmlValidator(
-      allowElementId: allowElementId,
-      allowClassName: allowClassName,
-    ).sanitize(htmlString);
-  }, zoneSpecification: ZoneSpecification(print: _printHandler));
+  return SaneHtmlValidator(
+    allowElementId: allowElementId,
+    allowClassName: allowClassName,
+  ).sanitize(htmlString);
 }
--- a/sanitize_html/lib/sanitize_html.dart
+++ b/sanitize_html/lib/sanitize_html.dart
@@ -15,7 +15,6 @@
 library sanitize_html;
 
 import 'dart:async' show Zone, ZoneDelegate, runZoned, ZoneSpecification;
-import 'package:universal_html/html.dart' as html;
 import 'src/sane_html_validator.dart' show SaneHtmlValidator;
 
 void _printHandler(Zone self, ZoneDelegate parent, Zone zone, String line) {
@@ -61,14 +60,10 @@ String sanitizeHtml(
   bool Function(String) allowElementId,
   bool Function(String) allowClassName,
 }) {
-  final doc = runZoned(() {
-    return html.DocumentFragment.html(
-      htmlString,
-      validator: SaneHtmlValidator(
-        allowElementId: allowElementId,
-        allowClassName: allowClassName,
-      ),
-    );
+  return runZoned(() {
+    return SaneHtmlValidator(
+      allowElementId: allowElementId,
+      allowClassName: allowClassName,
+    ).sanitize(htmlString);
   }, zoneSpecification: ZoneSpecification(print: _printHandler));
-  return doc.innerHtml;
 }
--- a/sanitize_html/lib/src/html_formatter.dart
+++ b/sanitize_html/lib/src/html_formatter.dart
@@ -1,0 +1,67 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+
+import 'package:html/dom.dart';
+
+final _attrEscape = HtmlEscape(HtmlEscapeMode.attribute);
+final _textEscape = HtmlEscape(HtmlEscapeMode.unknown);
+
+String formatHtmlNode(Node node) {
+  return _HtmlFormatter()._format(node);
+}
+
+class _HtmlFormatter {
+  final _sb = StringBuffer();
+
+  String _format(Node node) {
+    _writeNodes([node], 0);
+    return _sb.toString();
+  }
+
+  void _writeNodes(List<Node> nodes, int level) {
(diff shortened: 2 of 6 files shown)


sanitize_html/lib/sanitize_html.dart
```diff
@@ -61,14 +60,10 @@ String sanitizeHtml(
...
 }) {
-  final doc = runZoned(() {
-    return html.DocumentFragment.html(
-      htmlString,
-      validator: SaneHtmlValidator(
-        allowElementId: allowElementId,
-        allowClassName: allowClassName,
-      ),
-    );
+  return runZoned(() {
```
jonasfj: Yeah, we can remove `runZoned`, it was only used to deal with some print statements in `package:universial_html`

sanitize_html/lib/src/html_formatter.dart
```diff
@@ -0,0 +1,67 @@
...
+      if (node is Element) {
+        _writeElement(node, level);
+      } else if (node is Text) {
+        _sb.write(_textEscape.convert(node.text));
+      } else if (node is Document) {
+        _writeNodes([node.documentElement], level);
+      } else if (node is DocumentFragment) {
+        _writeNodes(node.nodes, level);
+      } else {
+        throw UnimplementedError('Unknown node: ${node.runtimeType} ($node)');
```
jonasfj: Maybe `assert(false, 'unknown node type')` followed by `return;` would be better?

That way, a user will never be able to crash the app..  but we might ignore unknown node types in production.

sanitize_html/lib/src/html_formatter.dart
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+  }
+
+  void _writeElement(Element elem, int level) {
+    final tagName = elem.localName;
+    _sb.write('<$tagName');
```
jonasfj: We are confident that `tagName` doesn't need escaping right? because `package:html` is an HTML5 parser...

sanitize_html/lib/src/html_formatter.dart
```diff
@@ -0,0 +1,67 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+
+import 'package:html/dom.dart';
+
+final _attrEscape = HtmlEscape(HtmlEscapeMode.attribute);
+final _textEscape = HtmlEscape(HtmlEscapeMode.unknown);
```
jonasfj: shouldn't this be: https://api.dartlang.org/stable/2.3.1/dart-convert/HtmlEscapeMode/element-constant.html

sanitize_html/lib/src/html_formatter.dart
jonasfj: In fact implementors of `Node` is:
> Comment Document DocumentFragment DocumentType Element Text

See: https://pub.dev/documentation/html/latest/dom/Node-class.html

I don't think it's important to support `DocumentType` and `Comment`, but maybe it's good for completeness..

sanitize_html/lib/src/html_formatter.dart
```diff
@@ -0,0 +1,67 @@
...
+        _writeNodes([node.documentElement], level);
+      } else if (node is DocumentFragment) {
+        _writeNodes(node.nodes, level);
+      } else {
+        throw UnimplementedError('Unknown node: ${node.runtimeType} ($node)');
+      }
+    }
+  }
+
+  void _writeElement(Element elem, int level) {
```
jonasfj: Level isn't used, let's just remove it

sanitize_html/lib/src/sane_html_validator.dart
```diff
@@ -217,35 +220,53 @@ class SaneHtmlValidator implements html.NodeValidator {
...
-    }
-    if (_alwaysAllowedAttributes.contains(attribute)) {
-      return true;
-    }
-    // Special validators for id and class on all elements
-    if (attribute == 'id') {
-      return allowElementId(element.id);
+  String sanitize(String htmlString) {
+    final root = html_parser.parseFragment(htmlString);
+    _sanitize(root);
```
jonasfj: Please `assert(root is Element)`, if this is not the case we're in big trouble.

Maybe if you start with `<html></html>` it won't be? or what about `<-- bla bla--->`... We'll probably need a few test cases for this as top-level.

sanitize_html/lib/src/sane_html_validator.dart
```diff
@@ -217,35 +220,53 @@ class SaneHtmlValidator implements html.NodeValidator {
...
+          return node.classes.isEmpty;
+        }
+        return !_isAttributeAllowed(tagName, attrName, v);
+      });
     }
-    if (attribute == 'class') {
-      return element.classes.every(allowClassName);
+    if (node.hasChildNodes()) {
+      for (int i = node.nodes.length - 1; i >= 0; i--) {
+        _sanitize(node.nodes[i]);
```
jonasfj: Maybe leave a comment that we do this in reverse, because we could otherwise skip one, when a node is removed...

sanitize_html/lib/src/sane_html_validator.dart
jonasfj: What if the htmlString is `<span></span><span></span>` won't the `root` be a `DocumentFragment`?

sanitize_html/lib/src/sane_html_validator.dart
```diff
@@ -217,35 +220,53 @@ class SaneHtmlValidator implements html.NodeValidator {
...
-      return allowElementId(element.id);
+  String sanitize(String htmlString) {
+    final root = html_parser.parseFragment(htmlString);
+    _sanitize(root);
+    return formatHtmlNode(root);
+  }
+
+  void _sanitize(Node node) {
+    if (node is Element) {
+      final tagName = node.localName.toUpperCase();
```
jonasfj: I wonder if it's possible to play tricks with `xmlns` to do things we don't want? Maybe we should filter everything that has:
https://pub.dev/documentation/html/latest/dom/Element/namespaceUri.html

I'm not sure what the test case would look like :)

jonasfj: Uh, maybe we don't need to worry about this because the `xmlns` attribute is not allowed..

sanitize_html/lib/src/html_formatter.dart
isoos: We could escape it like it were a text content... along with the attribute names?

sanitize_html/lib/src/sane_html_validator.dart
isoos: `parseFragment()` returns with a `DocumentFragment`, not `Element`, and `<span></span><span></span>` is already handled by that. I've added a test case to handle multiple top-level elements.

sanitize_html/lib/src/sane_html_validator.dart
isoos: As we ignore namespaces, both at processing and while outputting the formatted html, I think it is fair to say that tricks won't work...

isoos:
PTAL.

sanitize_html/lib/src/html_formatter.dart
jonasfj: How about `assert(false)`, instead of throwing `UnimplementedError`, because `assert` will only cause crashes in debug mode, where as `UnimplementedError` will cause crashes in production...

sanitize_html/lib/src/html_formatter.dart
jonasfj: I think we care confident, but maybe we should try to trick the parser and see if it'll produce sketchy tag names...

Oh, never mind, of course this is sane... We're sanitizing all the tag names and only allowing whitelisted tagnames. so How could they possibly have a tagName that needs to be escaped :)

sanitize_html/lib/src/html_formatter.dart
isoos: ack

isoos:
@jonasfj I don't have merge right, please merge this one.

PR #9, opened 2019-06-18
Travis fixes

kevmoo:


kevmoo:
Looks like this needs to be enabled – https://travis-ci.org/google/dart-neats

Issue #11, opened 2019-06-26
Only use self-closing tags for void-elements


jonasfj:
This is likely the cause of: https://github.com/dart-lang/pub-dev/issues/2405

For a list of void-element see: https://www.w3.org/TR/html5/syntax.html#void-elements

In particular:
> Tags are used to delimit the start and end of elements in the markup. Raw text, escapable raw text, and normal elements have a start tag to indicate where they begin, and an end tag to indicate where they end. The start and end tags of certain normal elements can be omitted, as described below in the section on [[#optional tags]]. Those that cannot be omitted must not be omitted. Void elements only have a start tag; end tags must not be specified for void elements. Foreign elements must either have a start tag and an end tag, or a start tag that is marked as self-closing, in which case they must not have an end tag.

It's clear that void-elements cannot have closing tags, but that for all other tags it is better to have `<b></b>` than `<b/>` as the latter is not allowed.

jonasfj:
This has been fixed and patch was released in `sanitize_html` version `1.3.0`, see:
https://pub.dev/packages/sanitize_html/versions/1.3.0#-changelog-tab-

PR #12, opened 2019-06-26
Fix issue #11, only use self-closing tags for void-elements

jonasfj:
Fixing #11


jonasfj:
@isoos, can you give this a review?

sanitize_html/lib/src/html_formatter.dart
```diff
@@ -19,6 +19,26 @@ import 'package:html/dom.dart';
 final _attrEscape = HtmlEscape(HtmlEscapeMode.attribute);
 final _textEscape = HtmlEscape(HtmlEscapeMode.element);
 
+// Set of HTML5 VOID-elements.
+//
+// See: https://www.w3.org/TR/html5/syntax.html#void-elements
+final _voidElements = Set.from([
```
isoos: nit: `<String>{}`

jonasfj: Good point :)

PR #13, opened 2019-07-02
Adding neat_cache

jonasfj:
I'ved tried to keep it simple... but please take your time, feel free to suggest new tests.

Testing socket issues is not pretty.


neat_cache/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
isoos: Fill these?

neat_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,149 @@
...
+      throw IntermittentCacheException('connect failed with exception');
+    }
+  }
+
+  Future<_RedisContext> _getContext() {
+    if (_context != null) {
+      return _context;
+    }
+    _context = _createContext();
+    scheduleMicrotask(() async {
```
isoos: Do you need this microtask here?

neat_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,149 @@
...
+        // If connecting fails, then we sleep and try again
+        await Future.delayed(_reconnectDelay);
+        _context = null; // reset _context, so next operation creates a new
+        return;
+      }
+      // If connecting was successful, then we await the connection being
+      // closed or error, and we reset _context.
+      // Any errors have already been returned if any commands were affect
+      // otherwise the error is just a broken TCP connection and we ignore it.
+      scheduleMicrotask(() async {
```
isoos: Do you need this microtask here?

neat_cache/test/redis_test.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  SocketProxyServer._(this._server, this._targetHost, this._targetPort);
+
+  static Future<SocketProxyServer> create(
+    String targetHost,
+    int targetPort,
+  ) async {
+    final server = await ServerSocket.bind('localhost', 0);
+    final sp = SocketProxyServer._(server, targetHost, targetPort);
+    scheduleMicrotask(() async {
```
isoos: this could be just `server.forEach(sp._process);` without the microtask wrapping, couldn't it? Or could server be a blocking mock impl?

neat_cache/LICENSE
jonasfj: No, this is just the:
> APPENDIX: How to apply the Apache License to your work

All of our license headers has a notice starting with "// Copyright 2019 Google LLC" :)

neat_cache/lib/src/providers/redis.dart
jonasfj: Good, point I'm already in a micro task, I don't need to create a new one :)

neat_cache/lib/src/providers/redis.dart
jonasfj: Yes, I'm returning before I'm awaiting...

1. If `_context` doesn't exist, I create it.
2. Then I return the `_context` future.
3. If I created `_context` in step (1), I then await it's failure, so that I can reset `_context = null` after a delay (if needed).

This ensures that we only make one connection at the time... no matter how many calls we have to `_getContext`.
Moreover, if the connection fails, we only do `_context = null` once. Until that has been done, all calls to `_getContext` will get a future that fails or a connection that is broken..

isoos: Wondering, not sure: would this be better if you'd use a `Completer`? The logic could be replace with checking the completer if it exists (if it does return its `.future`, if it does not, trigger the creation of the context).

neat_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,147 @@
...
+    }
+    _context = _createContext();
+    scheduleMicrotask(() async {
+      _RedisContext ctx;
+      try {
+        ctx = await _context;
+      } on IntermittentCacheException {
+        // If connecting fails, then we sleep and try again
+        await Future.delayed(_reconnectDelay);
+        _context = null; // reset _context, so next operation creates a new
```
isoos: This part is not really clear for me: what happens here? We had an exception, we reset a Future, but I'm no longer sure what the caller of this method will get, and what it is the relation with the current error handling.

jonasfj: We are awaiting the same future as the caller of the function gets.

So the caller of the function gets `IntermittentCacheException`.

In fact all callers of `_getContext` will get `IntermittentCacheException` until we've waited `_reconnectDelay` and then the next caller will get a new future, because `_context = null`.

neat_cache/lib/src/providers/redis.dart
jonasfj:
```dart
  Completer<_RedisContext> _completer;
  Future<_RedisContext> _getContext() {
    if (_completer != null) {
      return _completer.future;
    }
    _completer = Completer<_RedisContext>();
    scheduleMicrotask(() async {
      _RedisContext ctx;
      try {
        ctx = await _createContext();
        _completer.complete(ctx);
      } on IntermittentCacheException {
        _completer.completeError(IntermittentCacheException);
        // If connecting fails, then we sleep and try again
        await Future.delayed(_reconnectDelay);
        _completer = null; // reset _completer, so next operation creates a new
        return;
      } catch (e) {
        // This is not an IntermittentCacheException, we can't retry, we should
        // propagate the error, hopefully, this will crash the application.
        _completer.completeError(e);
      }
      // If connecting was successful, then we await the connection being
      // closed or error, and we reset _completer.
      // Any errors have already been returned if any commands were affect
      // otherwise the error is just a broken TCP connection and we ignore it.
      try {
        await ctx.connection.done;
      } catch (e) {
        // ignore error
      }
      _completer = null;
    });
    return _completer.future;
  }
```

This is the completer variant, it just feels more fragile to it that way... but it highlights some behavior, so we can consider if it's what we want..

neat_cache/test/redis_test.dart
jonasfj: Server could be mocking, but mocking redis is a lot of work...

And yeah, I didn't need to have a  microtask here, I added to avoid the lint that I'm not awaiting all futures :)

neat_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,154 @@
...
+        _reconnectDelay = reconnectDelay {
+    assert(_connectionString.isNotEmpty, 'connectionString must be given');
+    assert(!_connectTimeLimit.isNegative, 'connectTimeLimit is negative');
+    assert(!_commandTimeLimit.isNegative, 'commandTimeLimit is negative');
+    assert(!_reconnectDelay.isNegative, 'reconnectDelay is negative');
+  }
+
+  Future<_RedisContext> _createContext() async {
+    try {
+      print('### CONNECTING!!!');
```
isoos: Use logger instead.

neat_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,154 @@
...
+
+  Future<T> _withRedis<T>(_Op<T> fn) async {
+    if (_isClosed) {
+      throw StateError('CacheProvider.closed() has been called');
+    }
+    final ctx = await _getContext();
+    try {
+      return await fn(ctx.commands).timeout(_commandTimeLimit);
+    } on TimeoutException {
+      print('############ OP TIMED OUT');
```
isoos: Use logger instead.

neat_cache/lib/src/providers/redis.dart
isoos: Let's keep this current one.

neat_cache/lib/src/providers/redis.dart
jonasfj: These were 100% debug statements, for me when I was fixing the debugging the test... I thought I had removed them :see_no_evil:  :see_no_evil:  :see_no_evil:  :see_no_evil: :see_no_evil:

PR #18, opened 2019-08-09
Annotation and exception for API builder.

isoos:
First step to release `dart-lang/pub-dev/pkg/api_builder` :)

### DIFF
--- a/shelf_router/CHANGELOG.md
+++ b/shelf_router/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.7.2
+
+ * `EndPoint` annotation and `ApiResponseException` for API stub generation.
+
 ## v0.7.1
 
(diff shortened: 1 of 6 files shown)


jonasfj:
I'm not entirely sure we should release it just yet...

I think we're missing a few things:
 * JSON schemas for input/output
 * A sane way to annotate authentication
 * Generation of an [`openapi.json` reference](https://swagger.io/specification/).

Plus all sorts of sanity checks along the lines of:
 * Limit maximum input size,
 * Limit maximum output size,
 * Implicit support for HEAD requests whenever `GET` is implemented
 * Handling of `Accept` header
 * Handling of `Content-type` header
 * Maybe content encoding.

We have a lot of defensive work to do. It's also entirely possible that what we should do instead is to generate an abstract class from an open API reference document. That could be pretty neat.

PR #31, opened 2019-11-21
Initial release of package:safe_url_check

jonasfj:
Thinking we could use this to fix https://github.com/dart-lang/pana/issues/554

Note, I doubt requests to private IPs could actually cause any security vulnerabilities, but it is always nice to prevent these sorts of things.

### DIFF
--- a/safe_url_check/README.md
+++ b/safe_url_check/README.md
@@ -1,0 +1,43 @@
+`safe_url_check` for Dart
+=========================
+
+Utility to check if an untrusted URL is broken, without accidentally connecting
+to a private IP address.
+
+**Disclaimer:** This is not an officially supported Google product.
+
+When running in a Cloud environment a program usually has access to private IPv4
+addresses. This private IP-space might be used to grant access to database,
+caches, temporary credentials and various other services. If a program in such
+a cloud environment is checking untrusted URLs to see if a URL is broken, an
+attacker could fool the program into connecting to a private IP address by
+configuring DNS to resolve as such.
+
+This is generally undesirable. In most cases it is unlikely to cause any
+issues, as making a trivial `HEAD` or `GET` request to check if the URL is
+broken should be without side-effects. However, it's often preferable to harden
+security by protecting unauthorized access to the private IP space.
+
+This package offers a `safeUrlCheck` function, which makes a `HEAD` request and
+follows redirects after verifying that the host does not resolve to a private
+IPv4 address or locally unique IPv6 address.
+
+Note, it is plausible that it is desirable to restrict access to additional
+addresses space, pull-requests with suggestions are encouraged.
+
+## Example
+
+```dart
+import 'package:safe_url_check/safe_url_check.dart';
+
+Future<void> main() async {
+  // Check if https://google.com is a broken URL.
+  final exists = await safeUrlCheck(
+    Uri.parse('https://google.com'),
+    userAgent: 'myexample/1.0.0 (+https://example.com)',
+  );
+  if (exists) {
+    print('The url: https://google.com is NOT broken');
+  }
+}
+```
--- a/safe_url_check/lib/safe_url_check.dart
+++ b/safe_url_check/lib/safe_url_check.dart
@@ -65,7 +65,7 @@ Future<bool> safeUrlCheck(
   int maxRedirects = 8,
   String userAgent = _defaultUserAgent,
   HttpClient client,
-  RetryOptions retryOptions = const RetryOptions(),
+  RetryOptions retryOptions = const RetryOptions(maxAttempts: 3),
   Duration timeout = const Duration(seconds: 90),
 }) async {
   ArgumentError.checkNotNull(url, 'url');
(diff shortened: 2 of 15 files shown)


safe_url_check/lib/safe_url_check.dart
```diff
@@ -0,0 +1,182 @@
...
+/// following user-agent:
+/// ```
+/// User-Agent: package:safe_url_check/1.0.0 (+https://github.com/google/dart-neats/tree/master/safe_url_check)
+/// ```
+Future<bool> safeUrlCheck(
+  Uri url, {
+  int maxRedirects = 8,
+  String userAgent = _defaultUserAgent,
+  HttpClient client,
+  RetryOptions retryOptions = const RetryOptions(),
```
isoos: I think by default this should be a single try, and it should be the user's explicit setting to retry the URL multiple times.

safe_url_check/lib/safe_url_check.dart
```diff
@@ -0,0 +1,182 @@
...
+    if (500 <= response.statusCode && response.statusCode < 600) {
+      // retry again, when we hit a 5xx response
+      throw Exception('internal server error');
+    }
+    return response;
+  });
+  if (200 <= response.statusCode && response.statusCode < 300) {
+    return true;
+  }
+  if (response.isRedirect &&
```
isoos: Nice!

safe_url_check/lib/safe_url_check.dart
jonasfj: Maybe we can agree to `3` retries by default?

safe_url_check/README.md
```diff
@@ -0,0 +1,43 @@
...
+issues, as making a trivial `HEAD` or `GET` request to check if the URL is
+broken should be without side-effects. However, it's often preferable to harden
+security by protecting unauthorized access to the private IP space.
+
+This package offers a `safeUrlCheck` function, which makes a `HEAD` request and
+follows redirects after verifying that the host does not resolve to a private
+IPv4 address or locally unique IPv6 address.
+
+Note, it is plausible that it is desirable to restrict access to additional
+addresses space, pull-requests with suggestions are encouraged.
```
sigurdm: Do we have a guide somewhere telling us about this as best practice?
I just wonder, are there other what kinds of urls we would not like to access?

safe_url_check/lib/safe_url_check.dart
```diff
@@ -0,0 +1,182 @@
...
+/// ```
+/// User-Agent: package:safe_url_check/1.0.0 (+https://github.com/google/dart-neats/tree/master/safe_url_check)
+/// ```
+Future<bool> safeUrlCheck(
+  Uri url, {
+  int maxRedirects = 8,
+  String userAgent = _defaultUserAgent,
+  HttpClient client,
+  RetryOptions retryOptions = const RetryOptions(),
+  Duration timeout = const Duration(seconds: 90),
```
sigurdm: Sounds like a long default timeout.

safe_url_check/lib/safe_url_check.dart
```diff
@@ -0,0 +1,182 @@
...
+        return await _safeUrlCheck(
+          url,
+          maxRedirects,
+          c,
+          userAgent,
+          retryOptions,
+          timeout,
+        );
+      } finally {
+        c.close(force: true);
```
sigurdm: Nit: So you only close if you make your own client.
That is not obviously correct to me.
Should the absense of closing be documented?

Consider restructuring so there is only one call to _safeUrlCheck, and the closing happens conditionally.

safe_url_check/lib/safe_url_check.dart
```diff
@@ -0,0 +1,182 @@
...
+        isPrivateIpV4(ip) ||
+        isUniqueLocalIpV6(ip)) {
+      return false;
+    }
+  }
+
+  final response = await retryOptions.retry(() async {
+    // We can't use the HttpClient from dart:io with a custom socket, so instead
+    // of making a connection to one of the IPs resolved above, and specifying
+    // the host header, we reply on the OS caching DNS queries and not returning
```
sigurdm: reply -> rely

safe_url_check/lib/safe_url_check.dart
```diff
@@ -0,0 +1,182 @@
...
+        ip.isLinkLocal ||
+        ip.isMulticast ||
+        isPrivateIpV4(ip) ||
+        isUniqueLocalIpV6(ip)) {
+      return false;
+    }
+  }
+
+  final response = await retryOptions.retry(() async {
+    // We can't use the HttpClient from dart:io with a custom socket, so instead
```
sigurdm: What would that look like? Is there an open bug for that?

safe_url_check/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+homepage: https://github.com/google/dart-neats/tree/master/safe_url_check
+repository: https://github.com/google/dart-neats.git
+issue_tracker: https://github.com/google/dart-neats/labels/pkg:safe_url_check
+dependencies:
+  retry: ^3.0.0+1
+dev_dependencies:
+  test: ^1.5.1
+  pedantic: ^1.4.0
+  build_runner: ^1.7.2
+  build_version: ^2.0.1
```
sigurdm: I never saw this before.
We should use this in package:grpc!

safe_url_check/README.md
jonasfj: As this comment implies, there plausibly is other kinds of IPs that we don't want to access.

I reference RFCs above that justify blocking private IPv4 addresses. If you go reading through these RFCs you're likely to find more ip-blocks that could be blocked. I think loopback, link-local, and private ipv4 is the most sensitive ip-spaces, these are usually services on the same machine or on the same private virtual network (VPN).

Note: The risk of accessing these with a `GET` request should by definition be minimal, as `GET` is not allowed to have side effects.

safe_url_check/lib/safe_url_check.dart
jonasfj: Yes, this is also something you typically run in a background job.

safe_url_check/lib/safe_url_check.dart
jonasfj: I didn't check. I checked with `dart:io` owner that there is no way to do this.

We have it for `http2`, but I'm not sure how it would look in `dart:io`, and much less certain it could even be added without breaking compatibility.

PR #36, opened 2020-01-06
Fix package:shelf_router to ensure GET handlers also respond to HEAD …

### LINKED ISSUE #34: Using `router.get` should add a default handler for `router.head`
We should add default handlers for `HEAD` when we have handlers for `GET`.
Probably we can just serve the response from the `GET` handler if there is no other handler registered that handles the `HEAD` request.

```dart
// when writing
router.get('/', myHandler);
// we should implicitly add:
router.head('/', myHandler); // if and only if, no other registered handler matches the HEAD request.
```

@isoos, does this sound like a reasonable solution to you? I think this might solve some problems for us in `pub.dev`.

jonasfj:
…requests

Fixes #34

### DIFF
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -32,6 +32,17 @@ String params(Request request, String name) {
   return value;
 }
 
+/// Middleware to remove body from request.
+final _removeBody = createMiddleware(responseHandler: (r) {
+  if (r == null) {
+    return null;
+  }
+  if (r.headers.containsKey('content-length')) {
+    r = r.change(headers: {'content-length': '0'});
+  }
+  return r.change(body: <int>[]);
+});
+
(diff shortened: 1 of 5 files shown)


shelf_router/lib/src/router.dart
```diff
@@ -84,6 +95,11 @@ class Router {
     }
     verb = verb.toUpperCase();
 
+    if (verb == 'GET') {
```
isoos: What happens if you specify both `HEAD` and `GET`?

jonasfj: As long as you specify the `HEAD` handler before the `GET` handler you'll be fine.

isoos: This should be documented somewhere, otherwise this is good to go.

jonasfj: Good point, I documented it for `Router.get` but we should also document it here.

PR #46, opened 2020-06-03
fix retry overflow

wapa5pow:
When I use retry with relatively big number of attempt (like 64), delay is overflow. It is because delay is calculated by int.

I change it to floating point so that it won't overflow. I added test also.

### DIFF
--- a/retry/CHANGELOG.md
+++ b/retry/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v3.0.1+1
+## v3.0.1
  * Fix retry overflow
 
 ## v3.0.0+1
--- a/retry/CHANGELOG.md
+++ b/retry/CHANGELOG.md
@@ -1,3 +1,6 @@
+## v3.0.1+1
+ * Fix retry overflow
+
 ## v3.0.0+1
  * Update the example in `README.md`.
 
--- a/retry/lib/retry.dart
+++ b/retry/lib/retry.dart
@@ -104,7 +104,8 @@ class RetryOptions {
       return Duration.zero;
     }
     final rf = (randomizationFactor * (_rand.nextDouble() * 2 - 1) + 1);
-    final delay = (delayFactor * math.pow(2.0, attempt) * rf);
+    final exp = math.min(attempt, 31); // prevent overflows.
+    final delay = (delayFactor * math.pow(2.0, exp) * rf);
     return delay < maxDelay ? delay : maxDelay;
   }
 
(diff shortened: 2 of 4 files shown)


wapa5pow:
@jonasfj would you kindly review this?

retry/lib/retry.dart
```diff
@@ -104,7 +104,7 @@ class RetryOptions {
       return Duration.zero;
     }
     final rf = (randomizationFactor * (_rand.nextDouble() * 2 - 1) + 1);
-    final delay = (delayFactor * math.pow(2, attempt) * rf);
+    final delay = (delayFactor * math.pow(2.0, attempt) * rf);
```
jonasfj, line 107: Hmm, maybe it would be better to limit the exponent... 
In practice `maxDelay` should always make it less than `delayFactor * 2^31` (I hope :rofl:).

```suggestion
    final exp = math.max(attempt, 31); // prevent overflows.
    final delay = delayFactor * math.pow(2, exp) * rf;
```

jonasfj: This might just be a nit, but ints feels more predictable than floats here :D

wapa5pow: Your code make sense though `math.max` should be `math.min`.
I fix it.

jonasfj: Haha, well, at-least I was close, hehe :D

retry/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v3.0.1+1
```
jonasfj, line 1: Remove the `+1`

retry/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: retry
-version: 3.0.0+1
+version: 3.0.1+1
```
jonasfj, line 2: Remove the `+1`.

Context: version `3.0.0` is different from `3.0.0+1` ... the `+1` part is call _build metadata_, see: https://semver.org/#spec-item-10

It's optional, and I only used it because the fix I made was documentation-only. hence, it didn't change the code... just how it was packaged.

In practice I should probably just have made a patch release instead, since that's less confusing.

wapa5pow: I got your point! If I know about it, it make sense to add "+1" for documentation. thx.
I fix it.

jonasfj:
@wapa5pow great work, published https://pub.dev/packages/retry/versions/3.0.1

PR #54, opened 2020-08-08
Using runZoned to remove test pollution

### LINKED ISSUE #53: yaml_edit/test/fuzz_test.dart is very noisy
This test prints a lot of:
```
line 1, column 1: Warning: unknown directive.                                          
  ╷                                                                                                                                                                           
1 │ %*.*s                                                                                                                                                                     
  │ ^^^^^                                                                                                                                                                     
  ╵    
```
It's probably best if we can cleanup this noise, so that it doesn't pollute our test logs.

/cc @walnutdust

walnutdust:
Closes #53

### DIFF
--- a/yaml_edit/test/naughty_test.dart
+++ b/yaml_edit/test/naughty_test.dart
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:async';
+
 import 'package:test/test.dart';
 import 'package:yaml_edit/yaml_edit.dart';
 
@@ -21,11 +23,15 @@ void main() {
   for (final string in problemStrings) {
     test('expect string $string', () {
       final doc = YamlEditor('');
-
-      expect(() => doc.update([], string), returnsNormally);
-      final value = doc.parseAt([]).value;
-      expect(value, isA<String>());
-      expect(value, equals(string));
+      runZoned(() {
+        expect(() => doc.update([], string), returnsNormally);
+        final value = doc.parseAt([]).value;
+        expect(value, isA<String>());
+        expect(value, equals(string));
+      },
+          zoneSpecification: ZoneSpecification(
+              print: (Zone self, ZoneDelegate parent, Zone zone,
+                  String message) {}));
     });
   }
 }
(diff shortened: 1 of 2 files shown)


yaml_edit/test/naughty_test.dart
```diff
@@ -21,11 +23,15 @@ void main() {
...
-      expect(value, isA<String>());
-      expect(value, equals(string));
+      runZoned(() {
+        expect(() => doc.update([], string), returnsNormally);
+        final value = doc.parseAt([]).value;
+        expect(value, isA<String>());
+        expect(value, equals(string));
+      },
+          zoneSpecification: ZoneSpecification(
+              print: (Zone self, ZoneDelegate parent, Zone zone,
```
jonasfj, line 33: Can we leave a comment here, that we are doing this to throw away warnings from `package:yaml`.

yaml_edit/test/random_test.dart
```diff
@@ -54,8 +55,13 @@ dev_dependencies:
 ''');
 
       for (var j = 0; j < modificationsPerRound; j++) {
-        expect(
-            () => generator.performNextModification(editor), returnsNormally);
+        runZoned(() {
+          expect(
+              () => generator.performNextModification(editor), returnsNormally);
+        },
+            zoneSpecification: ZoneSpecification(
```
jonasfj, line 62: Again, let's add a comment that we're doing this hide warnings from `package:yaml`.

jonasfj: Also, can we run a single check that this won't eat errors / exceptions? (just a manual check will do, hehe)

From reading: https://api.dart.dev/stable/2.9.0/dart-async/runZoned.html
I don't think this is an error-zone, and only error-zones prevents errors from propagating to the calling context.

walnutdust:
Added comments in both suggested places, and tried adding both `throw ArgumentError.value(2);` and `expect(2, 3)` into the zone and they produced the expected output from `package:test` (i.e. errors/exceptions didn't seem to be eaten)!

jonasfj:
Let's merge when travis passes :D

PR #55, opened 2020-08-13
Added retryIfValue and onRetryIfValue to support retry on returned value condition

### LINKED ISSUE #45: Retry If Retruned Object Meets a Condition
Hi,

I'd like to retry if the returned object meets a condition.

Example:

```dart
      final response = await retry(
      // Make a GET request
      () => http.get('https://google.com'),
      // Retry on Exception
      retryIf: (e) => e is SocketException,
      retryIfReturnedObject: (o) => o == "retry-please"
  );
```
is this possible?
Thanks

ScottMacDougall:
Resolves #45

### DIFF
--- a/retry/lib/retry.dart
+++ b/retry/lib/retry.dart
@@ -44,6 +44,7 @@ final _rand = math.Random();
 ///   () => http.get('https://google.com').timeout(Duration(seconds: 5)),
 ///   // Retry on SocketException or TimeoutException
 ///   retryIf: (e) => e is SocketException || e is TimeoutException,
+///   retryIfValue: (v) => v == 500,
 /// );
 /// print(response.body);
 /// ```
@@ -110,9 +111,9 @@ class RetryOptions {
   }
 
   /// Call [fn] retrying so long as [retryIf] return `true` for the exception
-  /// thrown.
+  /// thrown, or if [retryIfValue] returns `true` for the value returned from [fn].
   ///
-  /// At every retry the [onRetry] function will be called (if given). The
+  /// At every retry the [onRetry] and [onRetryIfValue] functions will be called (if given). The
   /// function [fn] will be invoked at-most [this.attempts] times.
   ///
   /// If no [retryIf] function is given this will retry any for any [Exception]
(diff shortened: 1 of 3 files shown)


retry/lib/retry.dart
```diff
@@ -44,6 +44,7 @@ final _rand = math.Random();
 ///   () => http.get('https://google.com').timeout(Duration(seconds: 5)),
 ///   // Retry on SocketException or TimeoutException
 ///   retryIf: (e) => e is SocketException || e is TimeoutException,
+///   retryIfValue: (v) => v == 500,
```
jonasfj, line 47: This example won't work

retry/lib/retry.dart
```diff
@@ -122,13 +123,22 @@ class RetryOptions {
...
+    FutureOr<bool> Function(dynamic) retryIfValue,
+    FutureOr<void> Function(dynamic) onRetryIfValue,
   }) async {
     int attempt = 0;
     // ignore: literal_only_boolean_expressions
     while (true) {
       attempt++; // first invocation is the first attempt
       try {
-        return await fn();
+        var value = await fn();
```
jonasfj, line 134:
```suggestion
        final value = await fn();
```

retry/lib/retry.dart
```diff
@@ -122,13 +123,22 @@ class RetryOptions {
     FutureOr<T> Function() fn, {
     FutureOr<bool> Function(Exception) retryIf,
     FutureOr<void> Function(Exception) onRetry,
+    FutureOr<bool> Function(dynamic) retryIfValue,
+    FutureOr<void> Function(dynamic) onRetryIfValue,
```
jonasfj, line 127:
```suggestion
    FutureOr<void> Function(T) onRetryValue,
```
Why the if here? `onRetry` doesn't have that...

Also why are these `dynamic`, isn't `T` a better type? Will that work with `void`?

retry/lib/retry.dart
```diff
@@ -110,9 +111,9 @@ class RetryOptions {
   }
 
   /// Call [fn] retrying so long as [retryIf] return `true` for the exception
-  /// thrown.
+  /// thrown, or if [retryIfValue] returns `true` for the value returned from [fn].
   ///
-  /// At every retry the [onRetry] function will be called (if given). The
+  /// At every retry the [onRetry] and [onRetryIfValue] functions will be called (if given). The
```
jonasfj, line 116: Shouldn't be an OR, either `onRetry` is called or `onRetryValue` is called?

And don't we need to explain which is called when...

rajada1:
Any news ?

Bungeefan:
Wouldn't `retryIfResult` and `onRetryResult` be better names?
"Value" seems a bit generic.

Additionally, are there any plans on merging this PR?

retry/lib/retry.dart
```diff
@@ -122,13 +123,22 @@ class RetryOptions {
...
+    FutureOr<void> Function(dynamic) onRetryIfValue,
   }) async {
     int attempt = 0;
     // ignore: literal_only_boolean_expressions
     while (true) {
       attempt++; // first invocation is the first attempt
       try {
-        return await fn();
+        var value = await fn();
+        if (retryIfValue == null || !(await retryIfValue(value))) {
```
Bungeefan, line 135:
```suggestion
        if (attempt >= maxAttempts || retryIfValue == null || !(await retryIfValue(value))) {
```

Shouldn't there also be a `maxAttempts` check?
Otherwise, the users would have to count by themselves to not exceed the attempts if the value/result repeatedly triggers a retry. (e.g. HTTP 503 for longer than 30 seconds)

PR #56, opened 2020-08-18
package:tar

walnutdust:
Initial commit of package:tar. Currently only supports **reading** from TAR files, and supports the V7, USTAR, STAR, GNU, and PAX formats.

To-do:
1. Search up why TypeXGlobalHeaders were not supported in the golang implementation. - git blame doesn't yield anything


tar/lib/src/header.dart
```diff
@@ -0,0 +1,667 @@
...
+  /// Reference:
+  /// http://cdrtools.sourceforge.net/private/man/star/star.4.html
+  static const STAR = TarFormat._internal(16);
+}
+
+/// A [TarHeader] represents a single header in a tar archive.
+/// Some fields may not be populated.
+class TarHeader {
+  /// Type of header entry.
+  int typeFlag;
```
jonasfj, line 306: Maybe this should be an enum in Dart?

tar/lib/src/reader.dart
```diff
@@ -0,0 +1,424 @@
...
+import 'exceptions.dart';
+import 'header.dart';
+import 'stream.dart';
+import 'utils.dart';
+
+/// [TarReader] provides sequential access to the TAR files in a TAR archive.
+/// It is designed to read from a stream and to spit out substreams for
+/// individual file contents in order to minimize the amount of memory needed
+/// to read each archive where possible.
+class TarReader {
```
jonasfj, line 28: I am not entirely sure, so just brainstorming an alternative interface:
```dart

class TarEntry extends TarHeader {
  Stream<List<int>> contents;
  TarEntry._(header, contents) ...
}

Stream<TarEntry> readTarStream(Stream<List<int>> tarStream) async* {
  final r = TarReader(tarStream);
  while(r.next()) {
    yield TarEntry(r.header, r.contents);
  }
}
```

So then users would do:
```dart
await for(final entry in readTarStream(myTarStream)) {
  print(entry.name);
  await entry.drain();
}
```

Maybe it's a little crazy... because you in `TarReader` we can always make `TarReader.next()` either: drain the stream if not already drained, or throw a StateError if not drained.

tar/lib/src/header.dart
```diff
@@ -0,0 +1,667 @@
...
+  }
+  return TarFormat.V7;
+}
+
+/// Represents a [length]-sized fragment at [offset] in a file.
+///
+/// [SparseEntry]s can represent either data or holes, and we can easily
+/// convert between the two if we know the size of the file, all the sparse
+/// data and all the sparse entries combined must give the full size.
+class SparseEntry {
```
jonasfj, line 636: Does this need to be exported? I'm just wondering...

tar/lib/src/header.dart
```diff
@@ -0,0 +1,667 @@
...
+  /// 508   | 512 | Trailer - "tar\x00".
+  ///
+  /// Reference:
+  /// http://cdrtools.sourceforge.net/private/man/star/star.4.html
+  static const STAR = TarFormat._internal(16);
+}
+
+/// A [TarHeader] represents a single header in a tar archive.
+/// Some fields may not be populated.
+class TarHeader {
```
jonasfj, line 304: Let's make all exported classes `@sealed`

tar/lib/src/header.dart
walnutdust: nope - moved it and hid it!

tar/lib/src/reader.dart
walnutdust: It's kinda strange, and I still need some digging to understand why, but it seems that the current implementation doesn't require us to check and drain `_contents` - somehow `ChunkedStreamIterator.substream` handles this perfectly. 

either way, I think I should be able to convert to your alternative interface pretty easily - just let me know!

tar/lib/tar.dart
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Tape Archives (TAR) are a file format for storing a sequence of files that
+/// can be read and written in a streaming manner. This package aims to cover
+/// most variations of the format, including those produced by GNU and BSD tar
+/// tools.
+library tar;
+
+export 'src/exceptions.dart';
```
sigurdm, line 21:
```suggestion
export 'src/exceptions.dart' show TarFileException, TarHeaderException;
```

tar/lib/src/reader.dart
sigurdm: I also think that a stream of entries is slightly more "dart-y"

tar/lib/src/exceptions.dart
```diff
@@ -0,0 +1,39 @@
...
+
+/// Helper method to throw a [TarHeaderException].
+void headerException([String message = '']) {
+  throw TarHeaderException(message);
+}
+
+/// Helper method to throw a [TarFileException].
+void fileException([String message = '']) {
+  throw TarFileException(message);
+}
```
jonasfj, line 39: This kind of helper method won't make for a good stack trace.

But it might be reasonable to have a `createHeaderException([String message = ''])` method... that creates an instance of `TarHeaderException`.
Then we can make the constructor for `TarHeaderException` private, and we choose to not export `createHeaderException` in the library.

The only motivation for doing this is if we want to be able to change the constructor in the future, then we don't have to expose a public factory constructor.

tar/lib/src/exceptions.dart
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:meta/meta.dart';
+
+/// An exception indicating that there was an issue with parsing a
+/// TAR File. Intended to be seen by the user.
+@sealed
+class TarFileException extends FormatException {
+  TarFileException(String message) : super(message);
```
jonasfj, line 21: Private constructor, and public factory constructor prevents subclassing.

`@sealed` does not stop people, it just discourages them :D

tar/lib/src/header.dart
```diff
@@ -0,0 +1,367 @@
...
+        }
+      }
+
+      if (prefix.isNotEmpty) {
+        header.name = '$prefix/${header.name}';
+      }
+    }
+
+    if (!validateTypeFlag(header.typeFlag, header.format)) {
+      headerException('Invalid Header');
```
jonasfj, line 163: Using `throw createHeaderException...` would probably yield better stack traces :D

tar/lib/src/reader.dart
```diff
@@ -0,0 +1,442 @@
...
+@sealed
+class TarReader {
+  /// A chunked stream iterator to enable us to get our data 512 bytes at a time.
+  final ChunkedStreamIterator<int> _chunkedStream;
+
+  /// The [TarHeader] for the current file.
+  TarHeader header;
+
+  /// The current file contents.
+  Stream<List<int>> contents;
```
jonasfj, line 40: We should probably make this a getter... so that users can't assign to this :D

And maybe we should document in the comment how we expect people to read this..
Do we always expect them to drain it? What if they current entry is a directory? Can this be `null`.

It might be fair to say that if the current entry has contents then this must be drained.. but that if the current entry doesn't have contents draining it unnecessary (but allowed). What is the behavior?

tar/lib/src/reader.dart
```diff
@@ -0,0 +1,442 @@
...
+/// It is designed to read from a stream and to spit out substreams for
+/// individual file contents in order to minimize the amount of memory needed
+/// to read each archive where possible.
+@sealed
+class TarReader {
+  /// A chunked stream iterator to enable us to get our data 512 bytes at a time.
+  final ChunkedStreamIterator<int> _chunkedStream;
+
+  /// The [TarHeader] for the current file.
+  TarHeader header;
```
jonasfj, line 37: Maybe this should be a getter so users can't assign to it? :D

tar/lib/src/reader.dart
```diff
@@ -0,0 +1,442 @@
...
+      if (_skipNext > 0) {
+        await _chunkedStream.read(_skipNext);
+        _skipNext = 0;
+      }
+
+      /// The discarding of the remainder of the previous file should
+      /// already be handled by [ChunkedStreamIterator].
+      var rawHeader = await _chunkedStream.read(blockSize);
+
+      header = await _readHeader(rawHeader);
```
jonasfj, line 78: We might actually be updating the header in multiple steps..

I would suggest that while the future from `TarReader.next()` has not been completed, maybe calling the getter for `TarReader.header` should throw  a `StateError`. Rather than the behavior we have here where if you access `TarReader.header` while `await TarReader.next()` is still running, then you might be accessing an partially read header :see_no_evil:

jonasfj:
@walnutdust, I assume we'll have to fix the `ChunkedStreamIterator` bugs you've found before this will work reliably...

walnutdust:
haha yeah...

tar/lib/src/reader.dart
walnutdust: I opted to change it to a `Future` that completes right before `next` completes, in order to keep a behaviour that is somewhat like that of `StreamIterator` (i.e. `null` before the first `next` is called, and `null` when the stream is complete), while preventing users from reading it while `await TarReader.next()` is still running! 

Let me know if you think there are issues with this approach

tar/lib/src/reader.dart
walnutdust: Changed it to a getter!

Do you think I have to worry about the same issue as what you were commenting about `header`, that people while access it while `TarReader.next()` is still running?

walnutdust:
@jonasfj I ended up trying a different `ChunkedStreamIterator` implementation [here](https://github.com/google/dart-neats/pull/56/files#diff-dc791b386437ff238f4ec670c9e35bf1R63-R135) that seems to be playing nice.

With the way that it's implemented, I think `drain`-ing is now optional, and produces a good result in either case. I think not draining also won't cause an overload of system resources, and I've also tested for cases where there might be conflicts in `tar/test/stream_test.dart`. The output right now produces a `Stream<int>`, rather than a `Stream<List<int>>` though, and I'm not sure if that will pose any problem. 

I'll still want to see if I can get tests running on the Mac/Windows platforms, and maybe look into TypeXGlobalHeaders, but would appreciate feedback on the `ChunkedStreamIterator` implementation I have, and if there are any edge cases I need to test!

tar/lib/src/stream.dart
```diff
@@ -0,0 +1,135 @@
...
+  ///
+  /// The [ChunkedStreamIterator] is automatically cancelled if [read] goes
+  /// reaches the end of the stream or an error.
+  ///
+  /// Users should call [cancel] to ensure that the stream is properly closed
+  /// if they need to stop listening earlier than the end of the stream.
+  ///
+  /// Returns a future if the [cancel] operation is not completed
+  /// asynchronously. Otherwise returns null.
+  Future<dynamic> cancel() => _iterator.cancel();
```
jonasfj, line 109: I wuld avoid `Future<dynamic>` if there is not response the reader should care about just do: `Future<void>`

isoos (changes requested):
Only a drive-by high-level commentary...

tar/lib/src/stream.dart
```diff
@@ -0,0 +1,135 @@
...
+  final StreamIterator<T> _iterator;
+
+  /// Keeps track of the number of elements left in the current substream.
+  int _toRead = 0;
+
+  /// Keeps track of the current substream we are supporting.
+  int _substreamId = 0;
+
+  ChunkedStreamIterator(Stream<List<T>> stream)
+      : _iterator = StreamIterator(stream.expand((element) => element));
```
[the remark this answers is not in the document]
walnutdust: Thanks for the suggestion @isoos ! 

I think the primary goal we're trying to achieve is being able to read the input stream just once, without having to buffer the full contents of the file if we didn't want it. The use cases for this functionality are for example when we wish to only process certain file entries within a tar file while skipping the rest based on the headers, which allows us to save on memory.

You are right in that the current `ChunkedStreamIterator` as implemented here is inefficient, and I'm currently looking at removing the `stream.expand()` step and processing the stream in chunks, which should hopefully improve the space and time efficiency, and afterwards changing the constructor to a factory constructor that might return a more optimized iterator for `Uint8List`s by taking advantage of the `sublistView`.

I just took a look at `package:buffer`, and I'm not sure how I would use `package:buffer` to its fullest potential here. This `ChunkedStreamIterator` is really only meant to read and pass back chunks of raw bytes on demand. Could you advise me on how I might use the package to achieve the above goals?

/cc @jonasfj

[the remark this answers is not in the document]
isoos: > I think the primary goal we're trying to achieve is being able to read the input stream just once, without having to buffer the full contents of the file if we didn't want it.

Streams will buffer the events by default, as you have no control on when the Dart VM reads and forwards the bytes (or in how large chunks). The byte-stream coming from a `File` is an efficient 1-byte per value `Uint8List`, and leaving it in the Stream or buffering it via `ByteDataReader` is roughly the same overhead. However, `Stream.expand()` will blow up the memory requirement, as it is no longer a 1-byte `int` value, rather a 8-byte int value, and each value is copied to an expanded memory buffer.

For controlled byte-streams from files (and skipping ranges of bytes efficiently), you will need to use `RandomAccessFile`.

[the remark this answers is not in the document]
walnutdust: > Streams will buffer the events by default, as you have no control on when the Dart VM reads and forwards the bytes (or in how large chunks). 

hm, to my understanding, `File`'s `openRead()` is a single-subscription stream, and though I don't have precise control over when the Dart VM reads and forwards the bytes, wouldn't wrapping it in a `StreamIterator` pause the stream once the next element is processed (i.e. at the end of every `moveNext()`?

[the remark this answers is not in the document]
isoos: Not sure, maybe calling the `StreamSubscription.pause` and `.resume` in this case the VM does the right thing. We should test it with e.g. a 1-GB .tar file and see how the memory allocation profile goes.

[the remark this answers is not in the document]
walnutdust: Sorry about the delay, just got around to testing it. The `expand` really did make it super inefficient, so I modified the iterator to deal with chunks rather than expanding the elements out.

The testing was done on a 1-GB .tar file comprising two smaller files of 500 MB each, and the code used for testing can be found [here](https://gist.github.com/walnutdust/3fd79f9be2c0e324ba683d0bc93beff7). The testing code comprises of two parts, one where I ran through the file with the `TarReader` implemented with the `ChunkedStreamIterator` implementation, and another where we just did an `await for` loop through the elements of the file `Stream` doing no work to it. Time measured was the average of 100 iterations.

|Metric| TarReader | File Stream|
|----------|-----------|-----------|
|Average function run time| 341145.94 us | 337024.06 us |
|Peak Process Memory| 338.4MB | 328.7MB|
|New Generation Used:|1.5MB of 32.0MB | 1.5MB of 32.0MB|
|New Generation External| 21.5MB | 21.7MB|
|New Generation Collections| 748 | 748 |
|New Generation Average Collection Time| 0.37 ms| 0.35 ms|
|Old Generation Used| 6.4MB of 7.3MB| 5.7MB of 6.3MB|
|Old Generation External| 64B| 64B|
|Old Generation Collections| 4|3|
|Old Generation Average Collection Time| 1.78 ms| 2.46 ms|
|Total Generation Used| 7.9MB of 39.3MB| 7.2MB of 38.3MB|
|Total Generation External| 21.5MB| 21.7MB|
|Total Generation Collections| 752| 751|
|Total Generation Average Collection Time| 0.38 ms| 0.36 ms|

Naively looking at this, it appears to me that the `ChunkedStreamIterator` achieves our intended goals, and is pretty efficient about it. I am quite new to profiling, so please feel free to let me know if you want me to run more experiments with more parameters, or if there's something else that I should be analyzing!

tar/lib/src/utils.dart
```diff
@@ -0,0 +1,680 @@
...
+  RangeError.checkValidIndex(end - 1, bytes);
+
+  final rawCharCodes = bytes.sublist(start, end);
+
+  final stringEnd = rawCharCodes.indexOf(NUL);
+  final charCodes = rawCharCodes.sublist(0, stringEnd == -1 ? null : stringEnd);
+
+  try {
+    return utf8.decode(charCodes);
+  } catch (e) {
```
sigurdm, line 44: I believe we should only catch FormatErrors here.
```suggestion
  } on FormatError catch (e) {
```

sigurdm:
Not sure if character encoding of file names is handled correctly.

I tried making a folder called æøå, containing a file called æøå.dart. Then made a tar archive (mac os tar):
```
> tar cvf ææå.tar æøå
a æøå
a æøå/æøå.dart
```

Then I ran a small test:
```
import 'package:tar/tar.dart';
import 'dart:io';

main(List<String> args) async {
  final t = TarReader(File('æøå.tar').openRead());
  while (await t.next()) {
    print(await t.header);
  }
}
```
And got the output:

```
> dart bin/tar.dart
Unhandled exception:
FormatException: Non-ASCII characters detected in block
#0      new TarHeader (package:tar/src/header.dart:127:11)
#1      TarReader._readHeader (package:tar/src/reader.dart:199:12)
#2      TarReader.next (package:tar/src/reader.dart:92:26)
<asynchronous suspension>
#3      main (file:///Users/sigurdm/projects/blah/bin/tar.dart:6:18)
#4      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#5      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

Not sure if it is supposed to work - but the system tar can expand the archive just fine.

sigurdm:
For reference:
```
> base64 ææå.tar 
UGF4SGVhZGVyL8Omw7jDpQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADAwMDc1NSAANjY2NzY1IAAyNTc1MjMgADAwMDAwMDAwMDU2IDEzNzQxMzIyNzQwIDAxNzQ3MgAgeAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB1c3RhcgAwMHNpZ3VyZG0AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAcHJpbWFyeWdyb3VwAAAAAAAAAAAAAAAAAAAAAAAAAAAwMDAwMDAgADAwMDAwMCAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAxNiBwYXRoPcOmw7jDpS8KMzAgbXRpbWU9MTYwMjU5NDI3Mi42MDg2NjI5MTYKAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMOmw7jDpS8AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAwMDA3NTUgADY2Njc2NSAAMjU3NTIzIAAwMDAwMDAwMDAwMCAxMzc0MTMyMjc0MCAwMTU1NzIAIDUAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAdXN0YXIAMDBzaWd1cmRtAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHByaW1hcnlncm91cAAAAAAAAAAAAAAAAAAAAAAAAAAAMDAwMDAwIAAwMDAwMDAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAw6bDuMOlL1BheEhlYWRlci/DpsO4w6UuZGFydAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADAwMDY0NCAANjY2NzY1IAAyNTc1MjMgADAwMDAwMDAwMDcxIDEzNzQxMzIzMDEwIDAyMjU3NwAgeAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB1c3RhcgAwMHNpZ3VyZG0AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAcHJpbWFyeWdyb3VwAAAAAAAAAAAAAAAAAAAAAAAAAAAwMDAwMDAgADAwMDAwMCAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAyNyBwYXRoPcOmw7jDpS/DpsO4w6UuZGFydAozMCBtdGltZT0xNjAyNTk0MzEyLjEwNTg4NDMwOAoAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMOmw7jDpS/DpsO4w6UuZGFydAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAwMDA2NDQgADY2Njc2NSAAMjU3NTIzIAAwMDAwMDAwMjAwMCAxMzc0MTMyMzAxMCAwMjA2MjAAIDAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAdXN0YXIAMDBzaWd1cmRtAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHByaW1hcnlncm91cAAAAAAAAAAAAAAAAAAAAAAAAAAAMDAwMDAwIAAwMDAwMDAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
```

jonasfj:
Issues from `tar` on OS X might be reproducible using `bsdtar` from `sudo apt-get install libarchive-tools`.

walnutdust:
will look into it!

walnutdust:
@sigurdm @jonasfj The issue should be fixed - it was a misreading of the ustar specification which I thought would extend to the PAX format! I've added it as a test case too :)

tar/parse
```diff
@@ -0,0 +1,1062 @@
+TAR(1)                                      GNU TAR Manual                                      TAR(1)
```
jonasfj, line 1: Is this added by accident

tar/lib/src/reader.dart
```diff
@@ -0,0 +1,465 @@
...
+  /// A chunked stream iterator to enable us to get our data.
+  final ChunkedStreamIterator<int> _chunkedStream;
+
+  /// The [TarHeader] for the current file. Completes when a call to `next`
+  /// completes as well.
+  ///
+  /// Is `null` before the first call to `next` and after a call to `next`
+  /// completes with a `false` result.
+  Future<TarHeader> get header => _header;
+  Future<TarHeader> _header;
```
jonasfj, line 44: Should we consider making `header == null` while `next()` has not returned... once `next()` returns then we unassign this for `null`.

tar/lib/src/reader.dart
```diff
@@ -0,0 +1,465 @@
...
+@sealed
+class TarReader {
+  /// A chunked stream iterator to enable us to get our data.
+  final ChunkedStreamIterator<int> _chunkedStream;
+
+  /// The [TarHeader] for the current file. Completes when a call to `next`
+  /// completes as well.
+  ///
+  /// Is `null` before the first call to `next` and after a call to `next`
+  /// completes with a `false` result.
```
jonasfj, line 42: False result?

tar/parse
walnutdust: Yeap :fearful:

tar/lib/src/reader.dart
walnutdust: haha - completes with `false`. I tried updating the documentation, let me know if the new one works better or still suffers from the same issue.

tar/lib/src/reader.dart
walnutdust: hm - with the current implementation, header is a `Future` that is set to a new `Completer`'s future when `next` is called, then completed when `next()` returns. Making it `null` instead might make it less friendly, since currently the `Future` reminds them to `await` for the result, while if we changed it to `null`, they might use the `null` in other areas which leads to errors that might be harder to debug?

jonasfj: Well, you're supposed to be awaiting the `next()` call. So accessing `header` or `stream` before that completes seems unnecessary.

But maybe we should wait a bit, and see if Simon wants to take this somewhere.

simolus3: > But maybe we should wait a bit, and see if Simon wants to take this somewhere.

I do! I started to integrate the code for tar formats and header parsing from here (in https://github.com/simolus3/tar/commit/583c3398b46f6a974634f04697579e4e2bfc73af). I'll copy the remaining parser and tests soon.

jonasfj:
Some of this was published as [`package:tar`](https://pub.dev/packages/tar) huge credits to @simolus3 for shipping this.

For reference the repository is now: https://github.com/simolus3/tar/

Close this PR, thanks all!

PR #59, opened 2020-08-24
Updated Travis configuration to allow packages to be tested in parallel jobs

walnutdust:
- Added `acyclic_json`, `safe_url_check` and `yaml_edit` to travis jobs 
- Also commented out two `ensure_build_test` that were likely failing due to a `build_runner` issue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ env:
 jobs:
   include:
     - stage: analyze
-      name: `dartanalyzer .`
+      name: "dartanalyzer ."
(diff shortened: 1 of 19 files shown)


walnutdust:
If it's not super important to see the `analyze` stage as separate from the `test` stage, I can make the `script` key of Travis run `dartfmt` and `dartanalyzer` too, which gives us the ability to only have to update the main `env` variable, rather than in 3 places (main `env`, as well as the `env` for the analyze and format stages)

jonasfj:
> If it's not super important to see the analyze stage as separate from the test stage,

Yeah, that probably doesn't matter much :D

jonasfj (approved):

PR #65, opened 2020-09-07
Fixed ChunkedStreamIterator behavior when substream exactly completes a chunk.

### LINKED ISSUE #62: Bug in ChunkedStreamIterator.substream
When you call `ChunkedStreamIterator.substream`, if the first time the [wakeup function](https://github.com/google/dart-neats/blob/master/chunked_stream/lib/src/chunked_stream_iterator.dart#L174-L203) runs it manages to get `size` elements already, it resets `_wakeup()` to `null`. 

However, since the stream controller runs [`_subscription.resume`](https://github.com/google/dart-neats/blob/master/chunked_stream/lib/src/chunked_stream_iterator.dart#L132) when something subscribes to it, `_subscription` tries to [read in the next bit of data](https://github.com/google/dart-neats/blob/master/chunked_stream/lib/src/chunked_stream_iterator.dart#L39-L41) and finds a null `wakeup` but tries calling it anyways

The short test case below fails:

```dart
  test('substream ends with first chunk (fails)', () async {
    final s = ChunkedStreamIterator(_chunkedStream([
      ['a', 'b', 'c'],
      ['1', '2'],
    ]));
    expect(await s.read(1), equals(['a']));
    final i = s.substream(2);
    print(await i.length);
  });
```

walnutdust:
Closes #62

### DIFF
--- a/chunked_stream/CHANGELOG.md
+++ b/chunked_stream/CHANGELOG.md
@@ -1,10 +1,17 @@
+## v1.1.1
+
+- Fixed `substream` behavior when the input length exactly completes a chunk.
+
 ## v1.1.0
- * Added `asChunkedStream(N, input)` for wrapping a `Stream<T>` as a
-   chunked stream `Stream<List<T>>`, which is useful when batch processing
-   chunks of a stream.
+
+- Added `asChunkedStream(N, input)` for wrapping a `Stream<T>` as a
+  chunked stream `Stream<List<T>>`, which is useful when batch processing
+  chunks of a stream.
 
 ## v1.0.1
- * Fixed lints reported by pana.
+
+- Fixed lints reported by pana.
 
 ## v1.0.0
- * Initial release.
+
+- Initial release.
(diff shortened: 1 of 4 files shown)


jonasfj:
@walnutdust did this not work? :D

walnutdust:
@jonasfj ah I closed it because I realized there were plenty of test cases I hadn't considered, so I was not confident in the PR and was caught up in some stuff so I didn't have time to add test cases.

Will re-open after adding more test cases!

walnutdust:
I'm failing the following test:

```dart
  test('read() substream().cancel() read() -- one item at the time', () async {
    final s = ChunkedStreamIterator(_chunkedStream([
      ['a', 'b', 'c'],
      ['1', '2'],
    ]));
    expect(await s.read(1), equals(['a']));
    final i = StreamIterator(s.substream(3));
    await i.cancel();
    expect(await s.read(1), equals(['2']));
    expect(await s.read(1), equals([]));
  });
```
with or without the `_wakeup()` call in `substream`. Any idea/thoughts why? D:

jonasfj:
That's a good question... it's entirely possible that we need to reconsider the invariants in `ChunkedStreamIterator`.

Maybe we even need to write a few comments on why we have invariants the way we do :D

PR #69, opened 2020-09-14
RedisCacheProvider accept redis Connection or callback that returns a Connection

isoos:


neat_cache/lib/src/providers/redis.dart
```diff
@@ -27,36 +27,34 @@ class _RedisContext {
...
+  final RedisConnectionFn _connectionFn;
   final Duration _commandTimeLimit;
   final Duration _reconnectDelay;
 
   Future<_RedisContext> _context;
   bool _isClosed = false;
 
   RedisCacheProvider(
-    this._connectionString, {
+    /* String | Connection | RedisConnectionFn */ dynamic connection, {
```
jonasfj, line 42: Hehe, I think it's better to make two constructors.. But really, there is no need to worry about backwards compatibility here.

This class is not exported, it only exposed through static method `Cache.redisCacheProvider(String connectionString)` in:
https://pub.dev/documentation/neat_cache/latest/neat_cache/Cache/redisCacheProvider.html

I propose that we add a static method `Cache.redisCacheProviderFromConnection(FutureOr<Connection> Function(), ...)`.
This does add `Connection` from `package:dartis` to the public interface of this package. But that seems like an acceptable compromise.

We could also call it `Cache.redisCacheProvider2(...`, or maybe `Cache.dartisCacheProvider(FutureOr<Connection> Function(), ...)` because it expects a `Connection` object from `package:dartis`.

Hmm, thinking about it I suspect `Cache.dartisCacheProvider` might not be a bad name :D

jonasfj: side note: Let's just make it `FutureOr<Connection> Function()`, and not have `Connection` or `string` as possible values.. Because `RedisCacheProvider` generally does handle reconnecting.

jonasfj (changes requested):
note that `RedisCacheProvider` is not exposed, so we probably have to change the interface :D

neat_cache/lib/src/providers/redis.dart
isoos: > Cache.dartisCacheProvider

That would leak the `dartis` dependency at the top-level of the package, I'd rather have a `lib/dartis_cache_provider.dart`. wdyt?

jonasfj: `lib/dartis_cache_provider.dart` sounds like a reasonable solution...

jonasfj: Or maybe `lib/cache_providers/dartis.dart`, just to give it some structure.

jonasfj:
Closing for now, we can reopen later if we decide to make it easier to reuse connections.

Issue #71, opened 2020-10-14
addLinkRel not really optional when used from sanitizeHtml


ditman:
The new `addLinkRel` method is called [unconditionally](https://github.com/google/dart-neats/blob/master/sanitize_html/lib/src/sane_html_validator.dart#L253) but it might be null when calling the `sanitizeHtml` function directly, since it's optional according to [its interface](https://github.com/google/dart-neats/blob/master/sanitize_html/lib/sanitize_html.dart#L78).

This is [breaking things](https://github.com/flutter/flutter/issues/67854#issuecomment-707781063).

Both `allowElementId` and `allowClassName` do [null](https://github.com/google/dart-neats/blob/master/sanitize_html/lib/src/sane_html_validator.dart#L241) [checks](https://github.com/google/dart-neats/blob/master/sanitize_html/lib/src/sane_html_validator.dart#L244) before being used.

Example code:

```dart
sanitizeHtml('<a href="good-link">hello'); // boom
```

jonasfj:
Wow, obvious bug... Thanks for the PR.

ditman:
No problem, thanks for the package! :P

PR #72, opened 2020-10-14
[sanitize_html] Don't call addLinkRel if it's null

ditman:
(The title is fairly self-explanatory, but there's more detail in #71)

Fixes https://github.com/google/dart-neats/issues/71


jonasfj (approved):
r+

PR #74, opened 2020-10-16
Chunked Stream Iterator update

### LINKED ISSUE #62: Bug in ChunkedStreamIterator.substream

### LINKED ISSUE #66: Nested `ChunkedStreamIterator` propagates pausing
If we were to call `ChunkedStreamIterator` on a `ChunkedStreamIterator.substream`, the pausing propagates, leading to more pauses than expected on the parent `ChunkStreamIterator`.

```dart
  test('pause propagation', () async {
    final s = ChunkedStreamIterator(_chunkedStream([
      ['a', 'b', 'c'],
      ['1'],
      ['2'],
    ]));
    expect(await s.read(1), equals(['a']));
    final sub = s.substream(3);
    final c = ChunkedStreamIterator(sub);
    expect(await c.read(3), ['b', 'c', '1']);  /// Line (a)
    expect(await c.read(1), []); /// This line makes no difference to the result.
    expect(await s.read(1), equals(['2']));
  });
```

I think the following program flow is what causes the error:

1. In Line (a), the parent `ChunkedStreamIterator`, `s`, pauses after delivering the `['1']` block. 
1. At the same time, since `c` has delivered its full length, `c` will be `reset`. 
1. `reset` calls `_subscription.pause`, which calls `pause` in the substream controller for `sub`.
1. The controller for `sub` pauses the original stream subscription in `s`. 

The end result of this is that `s`'s stream subscription is paused multiple times, which causes future `read`s on `s` to fail because the subscription still remains paused.

walnutdust:
Updated the implementation of `ChunkedStreamIterator` to fix bugs raised in previous issues. As far as I can tell, the public API still stays the same, so it shouldn't be a breaking change I think?

Closes #66 
Closes #62 

/cc @jonasfj

### DIFF
--- a/chunked_stream/lib/src/chunked_stream_iterator.dart
+++ b/chunked_stream/lib/src/chunked_stream_iterator.dart
@@ -15,6 +15,8 @@
 import 'dart:async';
 import 'dart:math';
 
+import 'package:chunked_stream/src/read_chunked_stream.dart';
+
 /// Auxiliary class for iterating over the items in a chunked stream.
 ///
 /// A _chunked stream_ is a stream in which items arrives in chunks with each
@@ -68,6 +70,9 @@ class _ChunkedStreamIterator<T> implements ChunkedStreamIterator<T> {
   /// Keeps track of the current substream we are supporting.
   int _substreamId = 0;
 
+  /// Instance variable representing an empty list object, used as the empty
+  /// default state for [_buffered]. Take caution not to write code that
+  /// directly modify the [_buffered] list by adding elements to it.
   final List<T> _emptyList = [];
 
   /// Buffered items from a previous chunk. Items in this list should not have
@@ -90,7 +95,9 @@ class _ChunkedStreamIterator<T> implements ChunkedStreamIterator<T> {
   Future<List<T>> read(int size) async {
     /// Clears the remainder of elements if the user did not drain it.
     final readToIndex = min(_toRead, _buffered.length);
-    _buffered = _buffered.sublist(readToIndex);
+    _buffered = (readToIndex == _buffered.length)
+        ? _emptyList
+        : _buffered.sublist(readToIndex);
     _toRead -= readToIndex;
 
     while (_toRead > 0 && await _iterator.moveNext()) {
@@ -99,33 +106,11 @@ class _ChunkedStreamIterator<T> implements ChunkedStreamIterator<T> {
         _toRead = 0;
       } else {
         _toRead -= _iterator.current.length;
-        _buffered = [];
+        _buffered = _emptyList;
       }
     }
 
-    final chunks = <List<T>>[];
-
-    /// Starts by adding from the buffer if there are buffered elements
-    /// remaining.
-    if (_buffered.isNotEmpty) {
-      final addToIndex = min(size, _buffered.length);
-      chunks.add(_buffered.sublist(0, addToIndex));
-      _buffered = _buffered.sublist(addToIndex);
-      size -= addToIndex;
-    }
-
-    /// Grab as many chunks as needed and add them to [result], updating
-    /// [_buffered] as necessary.
-    while (size > 0 && await _iterator.moveNext()) {
-      _buffered = _iterator.current;
-
-      final addToIndex = min(size, _buffered.length);
-      chunks.add(_buffered.sublist(0, addToIndex));
-      _buffered = _buffered.sublist(addToIndex);
-      size -= addToIndex;
-    }
-
-    return chunks.expand((element) => element).toList();
+    return readChunkedStream(substream(size));
   }
 
   /// Cancels the stream iterator (and the underlying stream subscription)
--- a/chunked_stream/lib/src/chunked_stream_iterator.dart
+++ b/chunked_stream/lib/src/chunked_stream_iterator.dart
@@ -41,9 +41,6 @@ abstract class ChunkedStreamIterator<T> {
   /// Cancels the stream iterator (and the underlying stream subscription)
   /// early.
   ///
-  /// The [ChunkedStreamIterator] is automatically cancelled if [read] reaches
-  /// the end of the stream or an error.
-  ///
   /// Users should call [cancel] to ensure that the stream is properly closed
   /// if they need to stop listening earlier than the end of the stream.
   Future<void> cancel();
@@ -102,9 +99,6 @@ class _ChunkedStreamIterator<T> implements ChunkedStreamIterator<T> {
 
   /// Cancels the stream iterator (and the underlying stream subscription)
   /// early.
-  ///
-  /// The [ChunkedStreamIterator] is automatically cancelled if [read]
-  /// reaches the end of the stream or an error.
   ///
   /// Users should call [cancel] to ensure that the stream is properly closed
   /// if they need to stop listening earlier than the end of the stream.
--- a/chunked_stream/lib/src/chunked_stream_iterator.dart
+++ b/chunked_stream/lib/src/chunked_stream_iterator.dart
@@ -28,16 +28,6 @@ abstract class ChunkedStreamIterator<T> {
     return _ChunkedStreamIterator<T>(stream);
   }
 
-  /// Returns a list of the next [size] elements in original chunks. The size of
-  /// the chunks are not guaranteed to be consistent.
-  ///
-  /// Returns a list with a total of less than [size] elements split in chunks
-  /// if the end of stream is encounted before [size] elements are read.
-  ///
-  /// If an error is encountered before reading [size] elements, the error
-  /// will be thrown.
-  Future<List<List<T>>> readPreservingChunks(int size);
-
   /// Returns a list of the next [size] elements.
   ///
   /// Returns a list with less than [size] elements if the end of stream is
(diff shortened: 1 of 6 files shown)


chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +23,179 @@ import 'dart:math';
...
-      ChunkedStreamIterator._(stream);
+  /// Returns a list of the next [size] elements in original chunks. The size of
+  /// the chunks are not guaranteed to be consistent.
+  ///
+  /// Returns a list with a total of less than [size] elements split in chunks
+  /// if the end of stream is encounted before [size] elements are read.
+  ///
+  /// If an error is encountered before reading [size] elements, the error
+  /// will be thrown.
+  Future<List<List<T>>> readPreservingChunks(int size);
```
jonasfj, line 39: What is the motivation for this method? Do we need to expose it?

What is the behavior if I ask for `size = 7` and stream contains chunks of size 4? I'm not sure I can read that from the documentation.

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +23,179 @@ import 'dart:math';
...
 
-  /// Read a series of chunks from the stream until we have [size] number of
-  /// items, then return these items.
+  /// Cancels the stream iterator (and the underlying stream subscription)
+  /// early.
   ///
-  /// This returns less than [size], if end of stream occurs before [size]
-  /// number of items have been received.
+  /// The [ChunkedStreamIterator] is automatically cancelled if [read] or
+  /// [readPreservingChunks] reaches the end of the stream or an error.
```
jonasfj, line 54: What if `read` or `readPreservingChunks` throws an exception? Also why is this not documented in `read` or `readPreservingChunks`?

I think these semantics are fine.. I'm not sure it's normal or expected.. Honestly, I have a hard time grasping what is right. In most case this is probably preferable. We could add a boolean to the constructor `cancelOnError: true`.

Also what if an error happens while reading a `substream`? I almost expect the same behavior, I'm guessing it's just not documented.

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +23,179 @@ import 'dart:math';
...
-    }
-    if (_wakeup != null) {
-      throw StateError('Concurrent invocations not supported');
-    }
+  /// The resulting stream may contain less than [size] elements if the
+  /// underlying stream has less than [size] elements before the end of stream.
+  ///
+  /// If [read] or [readPreservingChunks] is called before the sub-[Stream] is
+  /// fully read, the remainder of the elements in the sub-[Stream] will be
+  /// automatically drained.
```
jonasfj, line 67: Would it not be preferable to make this an error?

We already state that the API cannot be called concurrently. I think it would be fair to elaborate that calling `read` or `substream` before a previously returned substream has been drained, consumed or cancelled throws.

IMO, we should throw a `StateError` in this case.

-----------------

I think we should consider making this an error rather than a silent draining, imagine: I'm reading a tar-file... I pass the substream off to something that sends over the network... But for some reason I forget to await that substream is fully sent before I read the next entry from the tar-reader.. Depending on how events a propagated and how fast my loop gets back to reading the next entry, it's possible that the thing sending my substream will end up sending a partial substream, right? (because next time I try to read an entry, my previous substream is drained).

I suspect you might want to say that, if nobody has started listening on the substream before the next call to `read`, then it is drained. But even in this case, I suspect it's preferable and more predictable to throw a `StateError` and tell the user to `.drain()` or cancel their subscription, before they call `read` again.

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +23,179 @@ import 'dart:math';
...
+
+        _buffered = _iterator.current;
       }
-    };
-    _wakeup();
 
-    return c.stream;
-  }
+      final yieldBlockEnd = min(_toRead, _buffered.length);
+      final toYield = _buffered.sublist(0, yieldBlockEnd);
```
jonasfj, line 193: We can improve performance here with a simple special case:
```dart
final toYield = _toRead < _buffered.length ? _buffered.sublist(0, _toRead) : _buffered;
```

That way we don't make an unnecessary copy... we can just return the `_buffered` chunk, if there is room for it.

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +23,179 @@ import 'dart:math';
...
+      _buffered = _buffered.sublist(yieldBlockEnd);
```
jonasfj, line 194: We could special case `_toRead < _buffered.length` here too..
```dart
_buffered = _toRead < _buffered.length ? buffered.sublist(_toRead) : _emptyList;
```

Then have an instance field:
```dart
final _emptyList = <T>[];
```
That way we don't  create a new empty list object whenever we need one... (we never mutate the `_buffered` list, so we can just use a special object to represent the empty list). This might even be better than using `null`, because it allows the compiler to skip `null` checks, if we initialize `_buffered = _emptyList`.

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +23,179 @@ import 'dart:math';
...
+  /// encounted before [size] elements are read.
+  ///
+  /// If an error is encountered before reading [size] elements, the error
+  /// will be thrown.
+  @override
+  Future<List<T>> read(int size) async {
+    return (await readPreservingChunks(size))
+        .expand((element) => element)
+        .toList();
   }
```
jonasfj, line 151: I think this could just be:
```dart
Future<List<T>> read(int size) => readChunkedStream(substream(size));
```

using [`bufferChunkedStream`](https://pub.dev/documentation/chunked_stream/latest/chunked_stream/bufferChunkedStream.html) from `package:chunked_stream` -- unless I also made a bug in this function :see_no_evil:

chunked_stream/lib/src/chunked_stream_iterator.dart
walnutdust: hm I think you're right.

I started writing this method because I thought there might be instances where users of the library might know the total number of elements to receive in advance, but wish to preserve the chunks - say the original stream was matrix data, or a stream of input/parameters/output and the program was doing some form of online processing. I just took a look at the areas where I used this function in the tar code, and it turns out I only ended up using this function to skip through the stream and avoid the extra work involved in `read` aha.

I'll remove this function for now, and with this gone we should have feature parity with the previous version of `ChunkedStreamIterator`, and if it turns out that this might be helpful, it shouldn't be too hard to add it back in.

chunked_stream/lib/src/chunked_stream_iterator.dart
jonasfj: This is just a nit :D

walnutdust: hehe... modifying it according to your suggestion right now! 

I'll still have to keep some of the logic for clearing the remainder of the elements if there was a previous `substream` that the user forgot to close, but your suggestion really helped made the code more concise :+1:

jonasfj: Are we sure we don't want an unclosed (and uncancelled) substream to simply throw `StateError`, next time user calls `.substream()`?

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +24,207 @@ import 'dart:math';
...
 
-  /// Cancel reading the stream.
-  ///
-  /// This may not be called concurrently with [read] or [substream].
-  Future<void> cancel() async {
-    if (_wakeup != null) {
-      throw StateError('Concurrent invocations not supported');
+  /// Asynchronous generator implementation for [substream].
+  Stream<List<T>> _substream() async* {
+    /// Only yield when there are elements to be read.
```
jonasfj, line 196: nit: inline comments usually only use `//` not `///`

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -23,197 +24,207 @@ import 'dart:math';
...
+  int _toRead = 0;
+
+  /// Buffered items from a previous chunk. Items in this list should not have
+  /// been read by the user.
+  List<T> _buffered;
+
+  /// Instance variable representing an empty list object, used as the empty
+  /// default state for [_buffered]. Take caution not to write code that
+  /// directly modify the [_buffered] list by adding elements to it.
+  final List<T> _emptyList = [];
```
jonasfj, line 99: You migh be able to use a constant list... then writing to it is not possible.
```suggestion
  final List<T> _emptyList = const <T>[];
```

jonasfj (approved):

chunked_stream/lib/src/chunked_stream_iterator.dart
walnutdust: This produces an error :O 

```
Constant list literals can't include a type parameter as a type argument, such as 'T'.
Try replacing the type parameter with a different type.dart(invalid_type_argument_in_const_literal)
```

jonasfj: ah,

PR #78, opened 2020-10-29
Fixed bug with adding empty map

### LINKED ISSUE #77: yaml_edit issues with empty maps in block style
**Example**
```dart
void main() {
  test('block problems', () {
    final e = YamlEditor('');
    e.update(
      [],
      wrapAsYamlNode(
        {'key': {}},
        collectionStyle: CollectionStyle.BLOCK,
      ),
    );
    return e.toString();
  });
}
```
Causes:
```
  Assertion failed: (package:yaml_edit) Failed to produce valid YAML after modification.

  # YAML before edit:
  > 
  
  # YAML after edit:
  > key:
  > {}
  
  Please file an issue at:
  https://github.com/google/dart-neats/issues/new?labels=pkg%3Ayaml_edit%2C+pending-triage&template=yaml_edit.md
  
  package:yaml_edit/src/editor.dart 585:7   YamlEditor._performEdit
  package:yaml_edit/src/editor.dart 258:14  YamlEditor.update
```
Expected out:

```yaml
key: {}
```
_Block style_ should be used as far as possible.. but it's clearly not possible to use block style for encoding the empty map.

Alternatively, we should do:
```yaml
key:
  {}
```
This would also be acceptable.

/cc @walnutdust

walnutdust:
Resolves #77.

If this PR is accepted, I'm planning on improving the testing and make it a patch release!

/cc @jonasfj

### DIFF
--- a/yaml_edit/lib/src/strings.dart
+++ b/yaml_edit/lib/src/strings.dart
@@ -279,6 +279,8 @@ String yamlEncodeBlockString(
       final formattedValue =
           yamlEncodeBlockString(entry.value, newIndentation, lineEnding);
 
+      /// Empty collections are always encoded in flow-style, so new-line must
+      /// be avoided
       if (isCollection(entry.value) && !isEmpty(entry.value)) {
         return formattedKey + ':\n' + formattedValue;
       }
--- a/yaml_edit/lib/src/strings.dart
+++ b/yaml_edit/lib/src/strings.dart
@@ -279,7 +279,7 @@ String yamlEncodeBlockString(
       final formattedValue =
           yamlEncodeBlockString(entry.value, newIndentation, lineEnding);
 
-      if (isCollection(entry.value)) {
+      if (isCollection(entry.value) && !isEmpty(entry.value)) {
         return formattedKey + ':\n' + formattedValue;
       }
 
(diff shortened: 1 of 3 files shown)


yaml_edit/lib/src/strings.dart
```diff
@@ -279,7 +279,7 @@ String yamlEncodeBlockString(
       final formattedValue =
           yamlEncodeBlockString(entry.value, newIndentation, lineEnding);
 
-      if (isCollection(entry.value)) {
+      if (isCollection(entry.value) && !isEmpty(entry.value)) {
```
jonasfj, line 282: Can we add a comment that:
`// Empty collections are always encoded in flow-style, so new-line must be avoided`

I assume this is why we have this hack?

jonasfj (approved):

yaml_edit/lib/src/strings.dart
walnutdust: Yeap! and added!

jonasfj:
@walnutdust we should probably do a `CHANGELOG.md` update too... and port this fix in to `pub` as well.

I'm happy to do a patch release as soon as we have the changelog and version bump in.

PR #83, opened 2020-11-02
Migrate retry to null-safety

sigurdm:


retry/lib/retry.dart
```diff
@@ -120,13 +120,13 @@ class RetryOptions {
...
-    FutureOr<void> Function(Exception) onRetry,
+    FutureOr<bool> Function(Exception)? retryIf,
+    FutureOr<void> Function(Exception)? onRetry,
   }) async {
-    int attempt = 0;
+    var attempt = 0;
     // ignore: literal_only_boolean_expressions
     while (true) {
-      attempt++; // first invocation is the first attempt
+      attempt++; // first invocation is the first attempt†
```
jonasfj, line 129: What is up with `†` ?

retry/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: retry
-version: 3.0.1
+version: 3.0.1-null-safety
```
jonasfj, line 2:
```suggestion
version: 3.0.1-null-safety.0
```

retry/pubspec.yaml
```diff
@@ -10,4 +10,4 @@ dev_dependencies:
   test: ^1.5.1
   pedantic: ^1.4.0
 environment:
-  sdk: ">=2.0.0 <3.0.0"
+  sdk: ">=2.12.0-0.0.dev <3.0.0"
```
jonasfj, line 13:
```suggestion
  sdk: ">=2.12.0-0 <2.12.0"
```

jonasfj (changes requested):

retry/lib/retry.dart
sigurdm: Hmm - no clue where that came from...

sigurdm: Removed

jonasfj (approved):

retry/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: retry
-version: 3.0.1
+version: 3.0.1-null-safety.0
```
natebosch, line 2: This should be `4.0.0-nullsafety`

- Packages that aren't pinned by flutter should use a breaking change.
- I think there can't be a `-` in the portion of the version following the first `-`. @jakemac53 do you remember if that is the case?
- The existing published `3.0.1` takes precedence over this one.

PR #85, opened 2020-11-03
Seal classes in shelf_router and prepare 0.7.3

jonasfj:


sigurdm (approved):
LGTM

Only the title is a bit misleading. Perhaps something like
"Seal classes in shelf_router and prepare 0.7.3"

jonasfj:
@graphicbeacon I randomly stumbled into your video and figured it would be useful if package users could find it.

graphicbeacon:
Thank you @jonasfj

PR #86, opened 2020-11-12
Router: allow to mount a Handler

xvrh:
`Router.mount()` can take a `Handler` so we can have middlewares for sub routes.

And make Router to be a Handler to preserve backward compatibility.

### DIFF
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -134,8 +134,12 @@ class Router {
   }
 
   /// Get a [Handler] that will route incoming requests to registered handlers.
+  @Deprecated('The Router class is a Handler on its own')
   Handler get handler => call;
 
+  /// Route incoming requests to registered handlers.
+  ///
+  /// This method allows a Router instance to be a [Handler].
   Future<Response> call(Request request) async {
     // Note: this is a great place to optimize the implementation by building
     //       a trie for faster matching... left as an exercise for the reader :)
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -115,18 +115,17 @@ class Router {
     _routes.add(RouterEntry('ALL', route, handler));
   }
 
-  /// Mount a router below a prefix.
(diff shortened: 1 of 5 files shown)


shelf_router/lib/src/router.dart
```diff
@@ -115,18 +115,17 @@ class Router {
     _routes.add(RouterEntry('ALL', route, handler));
   }
 
-  /// Mount a router below a prefix.
+  /// Mount a handler below a prefix.
   ///
   /// In this case prefix may not contain any parameters, nor
-  void mount(String prefix, Router router) {
+  void mount(String prefix, Handler handler) {
```
jonasfj, line 121: You are arguing that this isn't a breaking change because you added `Router.call` ?

jonasfj (commented):
Off the top of my head, this might not be such a bad idea...

shelf_router/lib/src/router.dart
```diff
@@ -135,24 +134,24 @@ class Router {
   }
 
   /// Get a [Handler] that will route incoming requests to registered handlers.
-  Handler get handler {
+  Handler get handler => call;
```
jonasfj, line 137: Should we consider marking this as `@deprecated`? Advocating that the `Router` class is an `Handler` on its own?

shelf_router/lib/src/router.dart
xvrh: Yes with a `call` method, any `Router` becomes a `Handler` (since handler is just a function definition).

```
Handler handler = Router();
```

shelf_router/lib/src/router.dart
xvrh: Yes, I can do that.

jonasfj (approved):
If you want to write a `CHANGELOG` entry and bump to `0.7.4` in `pubspec.yaml`, I'll look into publishing this :D

xvrh:
Thanks for the review @jonasfj. I added the changelog entry and bumped the version.

jonasfj (approved):

PR #95, opened 2021-01-08
Support custom buffers in chunked_stream

simolus3:
Adds an optional `newBuffer` argument to methods managing an internal buffer. When reading chunks of byte streams, a buffer implementation backed by typed data can be used to improve memory efficiency.

### DIFF
--- a/chunked_stream/lib/src/buffer_factory.dart
+++ b/chunked_stream/lib/src/buffer_factory.dart
@@ -1,0 +1,28 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Signature of a function creating a buffer.
+///
+/// The buffer returned must be growable.
+///
+/// Depending on the input stream to buffer, a custom [BufferFactory] can be
+/// provided to improve memory efficiency. For instance, a [Uint8Buffer][uib]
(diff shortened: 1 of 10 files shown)


jonasfj:
Sorry, I didn't see this PR before I had written most of https://github.com/google/dart-neats/pull/96 :D

simolus3:
That will work too, thanks!

chunked_stream/lib/src/buffer_factory.dart
```diff
@@ -0,0 +1,28 @@
...
+/// Depending on the input stream to buffer, a custom [BufferFactory] can be
+/// provided to improve memory efficiency. For instance, a [Uint8Buffer][uib]
+/// from the `typed_data` package can be suitable when operating on byte-
+/// streams.
+///
+/// When no custom [BufferFactory] is provided, this package will use the
+/// default [List] implementation.
+///
+/// [uib]: https://pub.dev/documentation/typed_data/latest/typed_data.typed_buffers/Uint8Buffer-class.html
+typedef BufferFactory<T> = List<T> Function();
```
jonasfj, line 28: Hmm, I think [BytesBuilder](https://api.dart.dev/stable/2.10.4/dart-typed_data/BytesBuilder-class.html) is a more recent addition to the Dart SDK.

In general, it's also better to use `Uint8List` because there are special optimizations for it everywhere, or at-least I think there is... I'm sure `Uint8Buffer` uses `Uint8List` behind the scenes, but we probably don't need growable buffers in general, we just have one place where we are growing a buffer.

simolus3: Yeah, `Uint8Buffer` uses a `Uint8List` under the hood. One advantage of a general buffer function is that we can have efficient chunked streams of floats or larger ints too, without needing specialized implementations in `chunked_stream`. But that's not nearly as common as the `uint8` case, so focusing on that might be easier.

jonasfj: A benefit of `Uint8List` is that the compiler can make certain optimizations if it knows for a fact that an instance of `Uint8List` or `List<int>` is an instance of the object from `dart:typed_data`. That said, if the compiler can't see if the `Uint8List` is a custom implementation of `Uint8List` then it can't make so many optimizations.

PR #96, opened 2021-01-08
Added readByteStream and prepare 1.4.0-null-safe.0 release

jonasfj:
@simolus3

Given that we use `.sublist` in `ChunkedStreamIterator` we don't really need to do much to byte stream specially.
Well, at-least not if those byte streams contains `Uint8List` which many of the streams from `dart:io` do, even if the type is `Stream<List<int>>`. Besides, even if there is a `List<int>` instance in there, most of the chunks are probably not going to be split, as they just pass right through.

The only place where it's a bit annoying is `ChunkedStreamIterator.read`, but I figured we could just introduce an extension method `ChunkedStreamIterator.readBytes` which returns `Uint8List`. I doubt there are many special cases where you want to use a specific kind of List type, it pretty much only makes sense for byte streams (I hope).

Also added `readByteStream` to complement `readChunkedStream`.

### DIFF
--- a/chunked_stream/README.md
+++ b/chunked_stream/README.md
@@ -14,13 +14,35 @@ be very inefficient, as each byte would be passed as an individual event.
 Instead bytes arrives in chunks (`List<int>`) and the type of a byte stream
 is `Stream<List<int>>`.
 
-To make it easy to work with the chunk streams, such as `Stream<List<int>>`,
+For easily converting a byte stream `Stream<List<int>>` into a single byte
+buffer `Uint8List` (which implements `List<int>`) this package provides
+`readByteStream(stream, maxSize: 1024*1024)`, which conviniently takes an
+optional `maxSize` parameter to help avoid running out of memory.
+
+**Example**
+```dart
+import 'dart:io';
+import 'dart:convert';
+import 'package:chunked_stream/chunked_stream.dart';
+
+Future<void> main() async {
+  // Open README.md as a byte stream
+  Stream<List<int>> fileStream = File('README.md').openRead();
+
+  // Read all bytes from the stream
+  final Uint8List bytes = await readByteStream(fileStream);
+  
+  // Convert content to string using utf8 codec from dart:convert and print
+  print(utf8.decode(bytes));
+}
+```
+
+To make it easy to process chunked streams, such as `Stream<List<int>>`,
 this package provides `ChunkedStreamIterator` which allows you to specify how
 many elements you want, and buffer unconsumed elements, making it easy to work
 with chunked streams one element at the time.
 
-## Example
-
+**Example**
 ```dart
 final reader = ChunkedStreamIterator(File('my-file.txt').openRead());
 // While the reader has a next byte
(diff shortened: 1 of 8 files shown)


jonasfj:
@simolus3 if you have any concerns please leave a comment in the review :D

chunked_stream/README.md
```diff
@@ -14,13 +14,35 @@ be very inefficient, as each byte would be passed as an individual event.
 Instead bytes arrives in chunks (`List<int>`) and the type of a byte stream
 is `Stream<List<int>>`.
 
-To make it easy to work with the chunk streams, such as `Stream<List<int>>`,
+For easily converting a byte stream `Stream<List<int>>` into a single byte
+buffer `Uint8List` (which implements `List<int>`) this package provides
+`readByteStream(stream, maxSize: 1024*1024)`, which conviniently takes an
```
sigurdm, line 19: typo
```suggestion
`readByteStream(stream, maxSize: 1024*1024)`, which conveniently takes an
```

chunked_stream/lib/src/chunked_stream_iterator.dart
```diff
@@ -111,13 +114,8 @@ class _ChunkedStreamIterator<T> implements ChunkedStreamIterator<T> {
   /// If an error is encountered before reading [size] elements, the error
```
sigurdm, line 114:
```suggestion
  /// If an error is encountered before reading [size] elements, an error
```

jonasfj: it is actually "the error" that was encountered that will be thrown :D

chunked_stream/test/chunked_stream_iterator_test.dart
```diff
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 import 'dart:async';
+import 'dart:ffi';
```
sigurdm, line 16: ??

chunked_stream/test/read_chunked_stream_test.dart
```diff
@@ -24,4 +26,22 @@ void main() {
...
+    final s = (() async* {
+      yield [1, 2];
+      yield Uint8List.fromList([3]);
+      yield [4];
+    })();
+    expect(await readByteStream(s), equals([1, 2, 3, 4]));
+  });
+
+  test('readByteStream returns Uint8List', () async {
+    final s = (() async* {
```
sigurdm, line 40: Do you want to check that this also happens when the input does not consist of Uint8Lists?

sigurdm (approved):
LGTM

chunked_stream/pubspec.yaml
```diff
@@ -9,5 +9,6 @@ issue_tracker: https://github.com/google/dart-neats/labels/pkg:chunked_stream
 dev_dependencies:
   test: ^1.16.0-nullsafety
   pedantic: ^1.4.0
+  meta: ^1.3.0-nullsafety.6
```
simolus3, line 12: Considering that we use `@sealed` in `lib/`, this should probably be a regular dependency?

simolus3 (commented):
> @simolus3 if you have any concerns please leave a comment in the review :D

This looks like what I need, thanks!

chunked_stream/pubspec.yaml
jonasfj: absolutely, that was a mistake.. Good catch.

PR #107, opened 2021-03-04
yaml_edit nullsafety

asartalo:
Hi,

For this part, I tried to focus the changes on the tests first by fixing the null safety errors I found when running `dart analyze`. There were analyze errors on the lib directory but I tried to fix only the parts where the interfaces don't match the tests and a few low-hanging fruits.

I figured this will be a good place start to discuss what needs to change.

Tagging @jonasfj

### DIFF
--- a/yaml_edit/lib/src/editor.dart
+++ b/yaml_edit/lib/src/editor.dart
@@ -158,7 +158,7 @@ class YamlEditor {
   /// If [orElse] is omitted, it defaults to throwing a [ArgumentError].
   ///
   /// To get a default value when [path] does not point to a value in the
-  /// [YamlNode]-tree, simply pass `orElse: () => [YamlNode]`.
+  /// [YamlNode]-tree, simply pass `orElse: () => ...`.
   ///
   /// **Example:** (using orElse)
   /// ```dart
--- a/yaml_edit/lib/src/editor.dart
+++ b/yaml_edit/lib/src/editor.dart
@@ -158,14 +158,17 @@ class YamlEditor {
   ///
   /// If [orElse] is omitted, it defaults to throwing a [ArgumentError].
   ///
-  /// To get `null` when [path] does not point to a value in the [YamlNode]-tree,
-  /// simply pass `orElse: () => null`.
+  /// To get a default value when [path] does not point to a value in the
+  /// [YamlNode]-tree, simply pass `orElse: () => [YamlNode]`.
   ///
   /// **Example:** (using orElse)
   /// ```dart
   /// final myYamlEditor('{"key": "value"}');
-  /// final value = myYamlEditor.valueAt(['invalid', 'path'], orElse: () => null);
-  /// print(value) // null
+  /// final node = myYamlEditor.valueAt(
+  ///   ['invalid', 'path'],
+  ///   orElse: () => wrapAsYamlNode(null),
+  /// );
+  /// print(node.value); // null
   /// ```
   ///
   /// **Example:** (common usage)
@@ -197,7 +200,7 @@ class YamlEditor {
   /// print(newNode.value); // "YAML"
   /// print(node.value); // "YAML Ain't Markup Language"
   /// ```
-  YamlNode? parseAt(Iterable<Object?> path, {YamlNode? Function()? orElse}) {
+  YamlNode parseAt(Iterable<Object?> path, {YamlNode Function()? orElse}) {
     ArgumentError.checkNotNull(path, 'path');
 
     return _traverse(path, orElse: orElse);
@@ -463,21 +466,21 @@ class YamlEditor {
   ///
   /// If [checkAlias] is `true`, throw [AliasError] if an aliased node is
   /// encountered.
-  YamlNode? _traverse(Iterable<Object?> path,
-      {bool checkAlias = false, YamlNode? Function()? orElse}) {
+  YamlNode _traverse(Iterable<Object?> path,
+      {bool checkAlias = false, YamlNode Function()? orElse}) {
     ArgumentError.checkNotNull(path, 'path');
     ArgumentError.checkNotNull(checkAlias, 'checkAlias');
 
     if (path.isEmpty) return _contents;
 
-    YamlNode? currentNode = _contents;
+    var currentNode = _contents;
     final pathList = path.toList();
 
     for (var i = 0; i < pathList.length; i++) {
       final keyOrIndex = pathList[i];
 
       if (checkAlias && _aliases.contains(currentNode)) {
-        throw AliasError(path, currentNode!);
+        throw AliasError(path, currentNode);
       }
 
       if (currentNode is YamlList) {
@@ -499,21 +502,21 @@ class YamlEditor {
           if (_aliases.contains(keyNode)) throw AliasError(path, keyNode);
         }
 
-        currentNode = map.nodes[keyNode];
+        currentNode = map.nodes[keyNode]!;
       } else {
-        return _pathErrorOrElse(path, path.take(i + 1), currentNode!, orElse);
+        return _pathErrorOrElse(path, path.take(i + 1), currentNode, orElse);
       }
     }
 
-    if (checkAlias) _assertNoChildAlias(path, currentNode!);
+    if (checkAlias) _assertNoChildAlias(path, currentNode);
 
     return currentNode;
   }
 
   /// Throws a [PathError] if [orElse] is not provided, returns the result
   /// of invoking the [orElse] function otherwise.
-  YamlNode? _pathErrorOrElse(Iterable<Object?> path, Iterable<Object?> subPath,
-      YamlNode parent, YamlNode? Function()? orElse) {
+  YamlNode _pathErrorOrElse(Iterable<Object?> path, Iterable<Object?> subPath,
+      YamlNode parent, YamlNode Function()? orElse) {
     if (orElse == null) throw PathError(path, subPath, parent);
     return orElse();
   }
--- a/yaml_edit/lib/src/editor.dart
+++ b/yaml_edit/lib/src/editor.dart
@@ -197,7 +197,7 @@ class YamlEditor {
   /// print(newNode.value); // "YAML"
   /// print(node.value); // "YAML Ain't Markup Language"
   /// ```
-  YamlNode parseAt(Iterable<Object> path, {YamlNode Function() orElse}) {
+  YamlNode? parseAt(Iterable<Object> path, {YamlNode? Function()? orElse}) {
     ArgumentError.checkNotNull(path, 'path');
 
     return _traverse(path, orElse: orElse);
@@ -244,7 +244,7 @@ class YamlEditor {
   ///   - test
   ///   - 2
   /// ```
-  void update(Iterable<Object> path, Object value) {
+  void update(Iterable<Object?> path, Object? value) {
     ArgumentError.checkNotNull(path, 'path');
 
     final valueNode = wrapAsYamlNode(value);
@@ -481,14 +481,14 @@ class YamlEditor {
       }
 
       if (currentNode is YamlList) {
-        final list = currentNode as YamlList;
+        final list = currentNode;
         if (!isValidIndex(keyOrIndex, list.length)) {
           return _pathErrorOrElse(path, path.take(i + 1), list, orElse);
         }
 
         currentNode = list.nodes[keyOrIndex];
       } else if (currentNode is YamlMap) {
-        final map = currentNode as YamlMap;
+        final map = currentNode;
 
         if (!containsKey(map, keyOrIndex)) {
           return _pathErrorOrElse(path, path.take(i + 1), map, orElse);
@@ -513,7 +513,7 @@ class YamlEditor {
   /// Throws a [PathError] if [orElse] is not provided, returns the result
   /// of invoking the [orElse] function otherwise.
   YamlNode _pathErrorOrElse(Iterable<Object> path, Iterable<Object> subPath,
-      YamlNode parent, YamlNode Function() orElse) {
+      YamlNode parent, YamlNode? Function()? orElse) {
     if (orElse == null) throw PathError(path, subPath, parent);
     return orElse();
   }
--- a/yaml_edit/lib/src/editor.dart
+++ b/yaml_edit/lib/src/editor.dart
@@ -87,7 +87,7 @@ class YamlEditor {
   String _yaml;
 
   /// Root node of YAML AST.
-  YamlNode _contents;
+  late YamlNode _contents;
 
   /// Stores the list of nodes in [_contents] that are connected by aliases.
   ///
@@ -197,7 +197,7 @@ class YamlEditor {
   /// print(newNode.value); // "YAML"
   /// print(node.value); // "YAML Ain't Markup Language"
   /// ```
-  YamlNode? parseAt(Iterable<Object> path, {YamlNode? Function()? orElse}) {
+  YamlNode? parseAt(Iterable<Object?> path, {YamlNode? Function()? orElse}) {
     ArgumentError.checkNotNull(path, 'path');
 
     return _traverse(path, orElse: orElse);
@@ -266,7 +266,7 @@ class YamlEditor {
 
     if (parentNode is YamlList) {
       final expected = wrapAsYamlNode(
-        [...parentNode.nodes]..[keyOrIndex] = valueNode,
+        [...parentNode.nodes]..[keyOrIndex as int] = valueNode,
       );
 
       return _performEdit(updateInList(this, parentNode, keyOrIndex, valueNode),
@@ -294,7 +294,7 @@ class YamlEditor {
   /// final doc = YamlEditor('[0, 1]');
   /// doc.appendToList([], 2); // [0, 1, 2]
   /// ```
-  void appendToList(Iterable<Object> path, Object value) {
+  void appendToList(Iterable<Object?> path, Object? value) {
     ArgumentError.checkNotNull(path, 'path');
     final yamlList = _traverseToList(path);
 
@@ -311,7 +311,7 @@ class YamlEditor {
   /// final doc = YamlEditor('[1, 2]');
   /// doc.prependToList([], 0); // [0, 1, 2]
   /// ```
-  void prependToList(Iterable<Object> path, Object value) {
+  void prependToList(Iterable<Object?> path, Object? value) {
     ArgumentError.checkNotNull(path, 'path');
 
     insertIntoList(path, 0, value);
@@ -329,7 +329,7 @@ class YamlEditor {
   /// final doc = YamlEditor('[0, 2]');
   /// doc.insertIntoList([], 1, 1); // [0, 1, 2]
   /// ```
-  void insertIntoList(Iterable<Object> path, int index, Object value) {
+  void insertIntoList(Iterable<Object?> path, int index, Object? value) {
     ArgumentError.checkNotNull(path, 'path');
     final valueNode = wrapAsYamlNode(value);
 
@@ -360,8 +360,8 @@ class YamlEditor {
   /// doc.spliceList([], 1, 0, ['Feb']); // [Jan, Feb, March, April, June]
   /// doc.spliceList([], 4, 1, ['May']); // [Jan, Feb, March, April, May]
   /// ```
-  Iterable<YamlNode> spliceList(Iterable<Object> path, int index,
-      int deleteCount, Iterable<Object> values) {
+  Iterable<YamlNode> spliceList(Iterable<Object?> path, int index,
+      int deleteCount, Iterable<Object?> values) {
     ArgumentError.checkNotNull(path, 'path');
     ArgumentError.checkNotNull(index, 'index');
     ArgumentError.checkNotNull(deleteCount, 'deleteCount');
@@ -417,11 +417,11 @@ class YamlEditor {
   /// - 2 # comment 2
   /// '''
   /// ```
-  YamlNode remove(Iterable<Object> path) {
+  YamlNode? remove(Iterable<Object?> path) {
     ArgumentError.checkNotNull(path, 'path');
 
-    SourceEdit edit;
-    YamlNode expectedNode;
+    SourceEdit? edit;
+    YamlNode? expectedNode;
     final nodeToRemove = _traverse(path, checkAlias: true);
 
     if (path.isEmpty) {
@@ -438,7 +438,7 @@ class YamlEditor {
     final parentNode = _traverse(collectionPath);
 
     if (parentNode is YamlList) {
-      edit = removeInList(this, parentNode, keyOrIndex);
+      edit = removeInList(this, parentNode, keyOrIndex as int);
       expectedNode = wrapAsYamlNode(
         [...parentNode.nodes]..removeAt(keyOrIndex),
       );
@@ -449,7 +449,7 @@ class YamlEditor {
           updatedYamlMap(parentNode, (nodes) => nodes.remove(keyOrIndex));
     }
 
-    _performEdit(edit, collectionPath, expectedNode);
+    _performEdit(edit ?? SourceEdit(0, 0, ''), collectionPath, expectedNode);
 
     return nodeToRemove;
   }
@@ -463,21 +463,21 @@ class YamlEditor {
   ///
   /// If [checkAlias] is `true`, throw [AliasError] if an aliased node is
   /// encountered.
-  YamlNode _traverse(Iterable<Object> path,
-      {bool checkAlias = false, YamlNode Function() orElse}) {
+  YamlNode? _traverse(Iterable<Object?> path,
+      {bool checkAlias = false, YamlNode? Function()? orElse}) {
     ArgumentError.checkNotNull(path, 'path');
     ArgumentError.checkNotNull(checkAlias, 'checkAlias');
 
     if (path.isEmpty) return _contents;
 
-    var currentNode = _contents;
+    YamlNode? currentNode = _contents;
     final pathList = path.toList();
 
     for (var i = 0; i < pathList.length; i++) {
       final keyOrIndex = pathList[i];
 
       if (checkAlias && _aliases.contains(currentNode)) {
-        throw AliasError(path, currentNode);
+        throw AliasError(path, currentNode!);
       }
 
       if (currentNode is YamlList) {
@@ -486,7 +486,7 @@ class YamlEditor {
           return _pathErrorOrElse(path, path.take(i + 1), list, orElse);
         }
 
-        currentNode = list.nodes[keyOrIndex];
+        currentNode = list.nodes[keyOrIndex as int];
       } else if (currentNode is YamlMap) {
         final map = currentNode;
 
@@ -501,25 +501,25 @@ class YamlEditor {
 
         currentNode = map.nodes[keyNode];
       } else {
-        return _pathErrorOrElse(path, path.take(i + 1), currentNode, orElse);
+        return _pathErrorOrElse(path, path.take(i + 1), currentNode!, orElse);
       }
     }
 
-    if (checkAlias) _assertNoChildAlias(path, currentNode);
+    if (checkAlias) _assertNoChildAlias(path, currentNode!);
 
     return currentNode;
   }
 
   /// Throws a [PathError] if [orElse] is not provided, returns the result
   /// of invoking the [orElse] function otherwise.
-  YamlNode _pathErrorOrElse(Iterable<Object> path, Iterable<Object> subPath,
+  YamlNode? _pathErrorOrElse(Iterable<Object?> path, Iterable<Object?> subPath,
       YamlNode parent, YamlNode? Function()? orElse) {
     if (orElse == null) throw PathError(path, subPath, parent);
     return orElse();
   }
 
   /// Asserts that [node] and none its children are aliases
-  void _assertNoChildAlias(Iterable<Object> path, [YamlNode node]) {
+  void _assertNoChildAlias(Iterable<Object?> path, [YamlNode? node]) {
     ArgumentError.checkNotNull(path, 'path');
 
     if (node == null) return _assertNoChildAlias(path, _traverse(path));
@@ -551,7 +551,7 @@ class YamlEditor {
   /// Throws [ArgumentError] if the element at the given path is not a
   /// [YamlList] or if the path is invalid. If [checkAlias] is `true`, and an
   /// aliased node is encountered along [path], an [AliasError] will be thrown.
-  YamlList _traverseToList(Iterable<Object> path, {bool checkAlias = false}) {
+  YamlList _traverseToList(Iterable<Object?> path, {bool checkAlias = false}) {
     ArgumentError.checkNotNull(path, 'path');
     ArgumentError.checkNotNull(checkAlias, 'checkAlias');
 
@@ -572,7 +572,7 @@ class YamlEditor {
   /// tree is equal to our expectations by deep equality of values. Throws an
   /// [AssertionError] if the two trees do not match.
   void _performEdit(
-      SourceEdit edit, Iterable<Object> path, YamlNode expectedNode) {
+      SourceEdit edit, Iterable<Object?> path, YamlNode? expectedNode) {
     ArgumentError.checkNotNull(edit, 'edit');
     ArgumentError.checkNotNull(path, 'path');
 
@@ -613,8 +613,8 @@ class YamlEditor {
   /// the whole tree.
   ///
   /// [SourceSpan]s in this new tree are not guaranteed to be accurate.
-  YamlNode _deepModify(YamlNode tree, Iterable<Object> path,
-      Iterable<Object> subPath, YamlNode expectedNode) {
+  YamlNode? _deepModify(YamlNode tree, Iterable<Object?> path,
+      Iterable<Object?> subPath, YamlNode? expectedNode) {
     ArgumentError.checkNotNull(path, 'path');
     ArgumentError.checkNotNull(tree, 'tree');
     RangeError.checkValueInInterval(subPath.length, 0, path.length);
@@ -628,11 +628,11 @@ class YamlEditor {
         throw PathError(path, subPath, tree);
       }
 
-      return wrapAsYamlNode([...tree.nodes]..[keyOrIndex] = _deepModify(
+      return wrapAsYamlNode([...tree.nodes]..[keyOrIndex as int] = _deepModify(
           tree.nodes[keyOrIndex],
           path,
           path.take(subPath.length + 1),
-          expectedNode));
+          expectedNode)!);
     }
 
     if (tree is YamlMap) {
--- a/yaml_edit/lib/src/editor.dart
+++ b/yaml_edit/lib/src/editor.dart
@@ -406,7 +406,7 @@ class YamlEditor {
   /// - 2 # comment 2
   /// '''
   /// ```
-  YamlNode? remove(Iterable<Object?> path) {
+  YamlNode remove(Iterable<Object?> path) {
     SourceEdit? edit;
     YamlNode? expectedNode;
     final nodeToRemove = _traverse(path, checkAlias: true);
@@ -414,7 +414,7 @@ class YamlEditor {
     if (path.isEmpty) {
       edit = SourceEdit(0, _yaml.length, '');
 
-      /// Parsing an empty YAML document returns `null`.
+      /// Parsing an empty YAML document returns YamlScalar with value `null`.
       _performEdit(edit, path, expectedNode);
       return nodeToRemove;
     }
--- a/yaml_edit/lib/src/editor.dart
+++ b/yaml_edit/lib/src/editor.dart
@@ -87,7 +87,7 @@ class YamlEditor {
   String _yaml;
 
   /// Root node of YAML AST.
-  late YamlNode _contents;
+  YamlNode _contents;
 
   /// Stores the list of nodes in [_contents] that are connected by aliases.
   ///
@@ -120,14 +120,12 @@ class YamlEditor {
 
   factory YamlEditor(String yaml) => YamlEditor._(yaml);
 
-  YamlEditor._(this._yaml) {
+  YamlEditor._(this._yaml) : _contents = loadYamlNode(_yaml) {
     _initialize();
   }
 
-  /// Loads [_contents] from [_yaml], and traverses the YAML tree formed to
-  /// detect alias nodes.
+  /// Traverses the YAML tree formed to detect alias nodes.
   void _initialize() {
-    _contents = loadYamlNode(_yaml);
     _aliases = {};
 
     /// Performs a DFS on [_contents] to detect alias nodes.
@@ -263,8 +261,11 @@ class YamlEditor {
     final parentNode = _traverse(collectionPath, checkAlias: true);
 
     if (parentNode is YamlList) {
+      if (keyOrIndex is! int) {
+        throw PathError(path, path, parentNode);
+      }
       final expected = wrapAsYamlNode(
-        [...parentNode.nodes]..[keyOrIndex as int] = valueNode,
+        [...parentNode.nodes]..[keyOrIndex] = valueNode,
       );
 
       return _performEdit(updateInList(this, parentNode, keyOrIndex, valueNode),
@@ -408,7 +409,7 @@ class YamlEditor {
   /// ```
   YamlNode remove(Iterable<Object?> path) {
     SourceEdit? edit;
-    YamlNode? expectedNode;
+    var expectedNode = wrapAsYamlNode(null);
     final nodeToRemove = _traverse(path, checkAlias: true);
 
     if (path.isEmpty) {
@@ -436,7 +437,7 @@ class YamlEditor {
           updatedYamlMap(parentNode, (nodes) => nodes.remove(keyOrIndex));
     }
 
-    _performEdit(edit ?? SourceEdit(0, 0, ''), collectionPath, expectedNode);
+    _performEdit(edit!, collectionPath, expectedNode);
 
     return nodeToRemove;
   }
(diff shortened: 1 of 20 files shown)


yaml_edit/lib/src/editor.dart
```diff
@@ -197,7 +197,7 @@ class YamlEditor {
   /// print(newNode.value); // "YAML"
   /// print(node.value); // "YAML Ain't Markup Language"
   /// ```
-  YamlNode parseAt(Iterable<Object> path, {YamlNode Function() orElse}) {
+  YamlNode? parseAt(Iterable<Object> path, {YamlNode? Function()? orElse}) {
```
asartalo, line 200: This change wasn't necessary to make the tests conform to null safety. However, I made the changes to conform to the documented interface saying:
```dart
  /// To get `null` when [path] does not point to a value in the [YamlNode]-tree,
  /// simply pass `orElse: () => null`.
```
This would mean that it's possible to get `null` as a result. When I did this change, some private methods needed to be changed to be null safe. I didn't change the private methods just so I can focus on the tests but this change made it necessary to change some of the tests to conform with the nullable result.

yaml_edit/lib/src/editor.dart
```diff
@@ -244,7 +244,7 @@ class YamlEditor {
   ///   - test
   ///   - 2
   /// ```
-  void update(Iterable<Object> path, Object value) {
+  void update(Iterable<Object?> path, Object? value) {
```
asartalo, line 247: This is the only interface change for `YamlEditor` that was necessary to conform with the tests on `test/special_test.dart`:

```dart
doc.update([null], 'tilde');
```
and on `test/update_test.dart` on line 839:
```dart
doc.update(['Mark McGwire'], null);
```

yaml_edit/test/test_case.dart
```diff
@@ -81,9 +81,9 @@ class _TestCase {
   final Uri goldenUri;
   final List<String> states = [];
 
-  String info;
-  YamlEditor yamlBuilder;
-  List<_YamlModification> modifications;
+  late String info;
+  late YamlEditor yamlBuilder;
+  late List<_YamlModification> modifications;
```
asartalo, line 86: Initialized later under `_initialize()`

yaml_edit/test/test_case.dart
```diff
@@ -225,9 +225,9 @@ Map _getValueFromYamlMap(YamlMap yamlMap) {
 dynamic _getValueFromYamlNode(YamlNode node) {
   switch (node.runtimeType) {
     case YamlList:
-      return _getValueFromYamlList(node);
+      return _getValueFromYamlList(node as YamlList);
     case YamlMap:
-      return _getValueFromYamlMap(node);
+      return _getValueFromYamlMap(node as YamlMap);
     default:
       return node.value;
```
asartalo, line 232: This whole section could have been changed to:
```dart
  if (node is YamlList) {
    return _getValueFromYamlList(node);
  }
  if (node is YamlMap) {
    return _getValueFromYamlMap(node);
  }
  return node.value;
```
...but I'm not sure of the implications of those changes yet.

yaml_edit/lib/src/editor.dart
jonasfj: Yeah, we're clearly using `null` as a sentinel value here.. Sometimes it's attractive to make a sentinel value, like a special instance of `YamlNode` that represents "no value". But I don't this is a good place..

The following is a value YAML document:
```yaml
key: value
otherkey: null
```
So we probably have to able to return null, because that could also be the value at the given path.

yaml_edit/lib/src/editor.dart
jonasfj: If we allow `null` in `path` arguments then we should do so for all methods that take a `path`.

Looking at the special test, I guess this is allowed because keys in YAML dictionaries can be `null`.

From the test case:
```dart
    final doc = YamlEditor('''
~: null # <-- The ~ character means null
false: false
No: No
true: true
''');
    doc.update([null], 'tilde');
    doc.update([false], false);
    doc.update(['No'], 'no');
    doc.update([true], 'true');

    expect(doc.toString(), equals('''
~: tilde # <-- The ~ character means null
false: false
No: no
true: "true"
'''));
```

Took a bit of googling to find out that [tilde is null in YAML](https://stackoverflow.com/questions/51990175/what-is-the-purpose-of-tilde-character-in-yaml).

Yes, YAML dictionaries don't just map from string to values, keys can be anything, even `null`. So I guess `null` is allowed in paths.

That probably also applied to `parseAt` and other methods that accept a `path`.

jonasfj: As for `doc.update(['Mark McGwire'], null);`, yes obviously, it's possible to set a key in YAML document to `null`. That makes sense.

Again, anywhere a value is inserted or replace we probably have to support inserting `null`. Like `appendToList` and `spliceList`, why shouldn't I be able to insert `null` into a list?

PS. feel free to add more special test cases to cover things we don't have tests for :D

jonasfj (commented):
I think it's a good idea to start with the public interface :D

Just watch out that we are consistent, many of the methods accept parameters like `path` or `value` that means the same in different operations..

asartalo (commented):
@jonasfj all tests now pass. I've updated some of the tests to make sure that null values and null paths are covered. I also had to modify `test/random_test.dart` to fix some errors when running them.

yaml_edit/lib/src/editor.dart
jonasfj: What happens today, if you call `yamlEditor.parseAt(['key1', 'key2'])` on a document like:
```yaml
key1:
  key2: null
```

Does it return a `YamlNode` with `YamlNode.value == null`, I think not, but please do test it.

yaml_edit/test/test_case.dart
jonasfj: `node is YamlList` means that `node` is an instance of some class that extends / implements `YamlList`.

However, `node.runtimeType == YamlList` only happens if `node` is an instance of  `YamlList`.

Look for other implementations of `YamlList` that might explain why this works this way.
Could also just be something obscure, you can try to change to test and see if it makes a difference.

If we refactor this, let's do it a separate PR (if it's null-safety unrelated).

yaml_edit/test/test_case.dart
```diff
@@ -237,12 +237,12 @@ dynamic _getValueFromYamlNode(YamlNode node) {
 /// objects.
 List<_YamlModification> _parseModifications(List<dynamic> modifications) {
   return modifications.map((mod) {
-    Object value;
-    int index;
-    int deleteCount;
+    Object? value;
+    var index = 0;
+    var deleteCount = 0;
```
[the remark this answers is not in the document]
jonasfj: Start by finding out if they can be null... read code and try to run the tests, maybe insert some print statements to help you debug it :D

yaml_edit/lib/src/editor.dart
jonasfj: @asartalo We need to find out of every place that takes or returns a `YamlNode` can return/take `null`.
I think the answer to that follows from `package:yaml` behavior, so I suggest exploring what that is.

yaml_edit/test/test_case.dart
asartalo: I see. Got it.

yaml_edit/test/append_test.dart
```diff
@@ -49,6 +49,28 @@ void main() {
...
+  - 0
+  - 1
+  - 2
+  - 3
+  - 4
+'''));
+      expectYamlBuilderValue(doc, {
+        null: [0, 1, 2, 3, 4]
+      });
+    });
```
asartalo, line 72: These are tests that explicitly cover null paths.

yaml_edit/test/append_test.dart
```diff
@@ -147,6 +169,13 @@ a:
       expectYamlBuilderValue(doc, [0, 1, 2, 3]);
     });
 
+    test('null value', () {
+      final doc = YamlEditor('[0, 1, 2]');
+      doc.appendToList([], null);
+      expect(doc.toString(), equals('[0, 1, 2, null]'));
+      expectYamlBuilderValue(doc, [0, 1, 2, null]);
+    });
+
```
asartalo, line 178: Tests that explicitly cover appending null values to list

yaml_edit/test/parse_test.dart
```diff
@@ -94,19 +94,28 @@ void main() {
...
+    });
+
+    test('with a null value in nested collection', () {
+      final doc = YamlEditor('''
+key1:
+  key2: null
+''');
+
+      expect(doc.parseAt(['key1', 'key2'])!.value, null);
     });
```
asartalo, line 119: Test for `parseAt()` with null value on a nested path.

yaml_edit/test/parse_test.dart
```diff
@@ -141,7 +150,17 @@ c: 3
...
+
+  test('works with null in path', () {
+    final doc = YamlEditor('{a: { ~: 4}}');
+    expect(doc.parseAt(['a', null])!.value, equals(4));
+  });
+
+  test('works with null value', () {
+    final doc = YamlEditor('{a: null}');
+    expect(doc.parseAt(['a'])!.value, equals(null));
+  });
```
asartalo, line 165: Test for parsing null values on a flow style doc.

yaml_edit/lib/src/editor.dart
asartalo: > What happens today, if you call yamlEditor.parseAt(['key1', 'key2']) on a document like:
>
> ```yaml
> key1:
>   key2: null
> ```
>
> Does it return a `YamlNode` with `YamlNode.value == null`, I think not, but please do test it.

@jonasfj Yes, right now on stable it does return a `YamlNode` with `YamlNode.value == null`. To be sure, I added some tests to demonstrate this on a clean master branch and found they pass. You can see this on PR #110 just so they are covered.

`YamlEditor.parseAt()` seems to guarantee to return `YamlNode` unless for exceptions (path errors, alias errors), or `null` when passed with `orElse: () => null` as described in the [documentation](https://pub.dev/documentation/yaml_edit/latest/yaml_edit/YamlEditor/parseAt.html). This is different from when we're accessing values directly using `package:yaml` like in the following:

```dart
import 'package:yaml/yaml.dart'; // yaml: ^3.1.0 (null safety)

void main(List<String> arguments) {
  const yamlSource = r'''
a:
  i: null
b:
  i: foo
''';

  final doc = loadYamlNode(yamlSource) as YamlMap;

  final value = doc['a']['i'];
  print('${value.runtimeType} - $value'); // Null - null

  final value2 = doc['b']['i'];
  print('${value2.runtimeType} - $value2'); // String - foo
}
``` 
In comparison, `YamlEditor.parseAt()` at current version (1.0.3), "always" returns a `YamlNode`.

```dart
import 'package:yaml_edit/yaml_edit.dart'; // yaml_edit: ^1.0.3

void main(List<String> arguments) {
  const yamlSource = r'''
a:
  i: null
b:
  i: foo
''';

  final doc = YamlEditor(yamlSource);

  final node = doc.parseAt(['a', 'i']);
  print('${node.runtimeType} - ${node.value}'); // YamlScalar - null

  final node2 = doc.parseAt(['b', 'i']);
  print('${node2.runtimeType} - ${node2.value}'); // YamlScalar - foo
}
```

`YamlEditor` achieves this by accessing the nodes directly.

The relevant lines are the following ([line 490-491 on master > editor.dart](https://github.com/google/dart-neats/blob/5ee7672e76b8303c6a33d002b5c1255dcc93ebf0/yaml_edit/lib/src/editor.dart#L490)):

```dart
      } else if (currentNode is YamlMap) {
        final map = currentNode as YamlMap;

        // ...snip...
        final keyNode = getKeyNode(map, keyOrIndex);

        // ...snip...
        currentNode = map.nodes[keyNode];
      } else {
```

In the above, `map.nodes` is of type `Map<dynamic, YamlNode>` and `currentNode` is always returned when found (except for `orElse` that returns null).

I've added some tests to cover this behavior for `YamlEditor.parseAt()` on my PR. If this is unclear, let me know.

jonasfj: Perfect, this is a good find! :smile: 

Let's make the signature: `YamlNode parseAt(Iterable<Object?> path, YamlNode Function() orElse)`
Then we still have `orElse`, but it has to return a `YamlNode`.

If we want, we can always add a `parseAtOrNull` function, but that's hardly very useful, since a user will likely have to check if the response is an `YamlScalar` with `.value == null` anyways.

yaml_edit/test/test_case.dart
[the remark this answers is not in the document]
jonasfj: This is just test code, right? So I'm not sure it matters all that much... So long as the tests are testing the code, you can use `dynamic` for all I care :rofl:  (not that using `dynamic` is advisable)

yaml_edit/test/parse_test.dart
jonasfj: Let's make `parseAt` always return `YamlNode` instead of `YamlNode?` -- as you've suggested, then we don't need `!` all over the code.

yaml_edit/lib/src/editor.dart
jonasfj: @asartalo  actually, can't we make the signature:
`YamlNode parseAt(Iterable<Object?> path, {YamlNode? Function() orElse?})`

Then, if `orElse` returns `null`, we just wrap it in a `YamlScalar`, and return an `YamlScalar` instance with `.value == null`?

That way, a user calling `parseAt(['key'], orElse: () => null)` will get the same result whether the document is:
```yaml
{key: null}   # key is present, but the value is "null"
```
OR:
```yaml
{}  # key is not present
```
Which is likely what the user actually wants.

jonasfj: Or we could make the signature:
`YamlNode parseAt(Iterable<Object?> path, {YamlNode Function() orElse?})`

And then simply update the documentation to not suggest:
```
final value = myYamlEditor.valueAt(['invalid', 'path'], orElse: () => null);
```
But instead suggest:
```
final value = myYamlEditor.valueAt(['invalid', 'path'], orElse: () => wrapAsYamlNode(null));
```

In fact, I think it's probably a mistake that we don't recommend that today, because the idea with `orElse: () => null` was that you wanted the same result regardless of whether the key was `null` or not present. This pattern of treating `null` and omitted key as the same is a very common and reasonable pattern :D

So I think you could argue that allowing `orElse` to return `null` invites bugs, because it makes people think they are handling the case where the value is at `path` in the YAML document is `null`, which is not what they are handling.

Always lovely to catch a bug when migrating to null-safety :rocket:

asartalo: I agree. It also one less codepath to remember as a user of the library and as a maintainer. I like the `YamlNode parseAt(Iterable<Object?> path, {YamlNode Function() orElse?})` signature . I'll add those changes to the documentation as well.

yaml_edit/test/test_case.dart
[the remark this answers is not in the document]
asartalo: Yes they are for testing. I'll compare test coverage between this change and on current stable just to be safe.

yaml_edit/lib/src/editor.dart
```diff
@@ -360,8 +360,8 @@ class YamlEditor {
   /// doc.spliceList([], 1, 0, ['Feb']); // [Jan, Feb, March, April, June]
   /// doc.spliceList([], 4, 1, ['May']); // [Jan, Feb, March, April, May]
   /// ```
-  Iterable<YamlNode> spliceList(Iterable<Object> path, int index,
-      int deleteCount, Iterable<Object> values) {
+  Iterable<YamlNode> spliceList(Iterable<Object?> path, int index,
+      int deleteCount, Iterable<Object?> values) {
     ArgumentError.checkNotNull(path, 'path');
     ArgumentError.checkNotNull(index, 'index');
     ArgumentError.checkNotNull(deleteCount, 'deleteCount');
```
asartalo, line 367: I would like to remove these `ArgumentError.checkNotNull()` calls as well since they're already handled by Dart.

yaml_edit/lib/src/editor.dart
```diff
@@ -417,11 +417,11 @@ class YamlEditor {
   /// - 2 # comment 2
   /// '''
   /// ```
-  YamlNode remove(Iterable<Object> path) {
+  YamlNode? remove(Iterable<Object?> path) {
```
asartalo, line 420: I also began looking at public methods where `YamlNode?` are returned as well. For `remove()`, the only documentation I can see about the `null` path is on the comment at line 430:

```dart
/// Parsing an empty YAML document returns `null`.
_performEdit(edit, path, expectedNode);
```

It also doesn't have tests that demonstrate using the return value of `remove()`. :(

asartalo: BTW, this was the only public method on editor that still returns `YamlNode?`. Everything else either is just `void` or something else and `Iterable<YamlNode>`.

yaml_edit/lib/src/editor.dart
```diff
@@ -121,7 +121,6 @@ class YamlEditor {
   factory YamlEditor(String yaml) => YamlEditor._(yaml);
 
   YamlEditor._(this._yaml) {
-    ArgumentError.checkNotNull(_yaml);
```
asartalo, line 124: I've removed all calls to `ArgumentError.checkNotNull()` since these should be handled by Dart itself.

yaml_edit/lib/src/editor.dart
```diff
@@ -417,17 +406,15 @@ class YamlEditor {
...
-    YamlNode expectedNode;
+  YamlNode remove(Iterable<Object?> path) {
+    SourceEdit? edit;
+    YamlNode? expectedNode;
     final nodeToRemove = _traverse(path, checkAlias: true);
 
     if (path.isEmpty) {
       edit = SourceEdit(0, _yaml.length, '');
 
-      /// Parsing an empty YAML document returns `null`.
```
asartalo, line 430: Updating comment for future maintainers

yaml_edit/test/remove_test.dart
```diff
@@ -48,6 +48,20 @@ c: 3
...
+      final node = doc.remove(['a', 'b']);
+      expect(node.value, equals('foo'));
+    });
+
+    test('returns null-value node when doc is empty and path is empty', () {
+      final doc = YamlEditor('');
+      final node = doc.remove([]);
+      expect(node.value, equals(null));
+    });
+  });
```
asartalo, line 63: Tests to cover `remove()` return values.

yaml_edit/lib/src/editor.dart
```diff
@@ -197,9 +199,7 @@ class YamlEditor {
   /// print(newNode.value); // "YAML"
   /// print(node.value); // "YAML Ain't Markup Language"
   /// ```
-  YamlNode parseAt(Iterable<Object> path, {YamlNode Function() orElse}) {
-    ArgumentError.checkNotNull(path, 'path');
-
+  YamlNode parseAt(Iterable<Object?> path, {YamlNode Function()? orElse}) {
```
asartalo, line 202: `parseAt()` now returns a non-nullable `YamlNode` and orElse must return the same.

yaml_edit/lib/src/editor.dart
```diff
@@ -158,14 +157,17 @@ class YamlEditor {
...
   /// **Example:** (using orElse)
   /// ```dart
   /// final myYamlEditor('{"key": "value"}');
-  /// final value = myYamlEditor.valueAt(['invalid', 'path'], orElse: () => null);
-  /// print(value) // null
+  /// final node = myYamlEditor.valueAt(
+  ///   ['invalid', 'path'],
+  ///   orElse: () => wrapAsYamlNode(null),
+  /// );
+  /// print(node.value); // null
```
asartalo, line 170: I've updated the documentation here to reflect the new behavior. Please let me know if this is enough. I'm not yet sure if I worded it correctly.

yaml_edit/test/append_test.dart
jonasfj: I think it's fair that `wrapAsYamlNode` can take null... and then it's also fair that `value` in `insertIntoList` (and friends), is null-able... because those values are automatically wrapped with `wrapAsYamlNode`.

yaml_edit/lib/src/editor.dart
jonasfj: Yes, this is one of the nice things we can now do :D

yaml_edit/test/remove_test.dart
jonasfj: Feel free to break additional tests into a separate PR.. those are easy to land independently :D

yaml_edit/lib/src/editor.dart
jonasfj: I think this is solid..

yaml_edit/lib/src/editor.dart
```diff
@@ -87,7 +87,7 @@ class YamlEditor {
   String _yaml;
 
   /// Root node of YAML AST.
-  YamlNode _contents;
+  late YamlNode _contents;
```
jonasfj, line 90: If we want to avoid `late` we can change the constructor to:
`YamlEditor._(this._yaml) : _contents = loadYamlNode(_yaml) {` 

And the update `_initialize` to not run `loadYamlNode`... then we would have to patch `_performEdit` which also calls `_initialize`, but that might actually be good because it runs `loadYamlNode` (once on it's own) and once through `_initialize`, so maybe that's good to cleanup.

We can also do this later...

yaml_edit/lib/src/editor.dart
```diff
@@ -158,14 +157,17 @@ class YamlEditor {
   ///
   /// If [orElse] is omitted, it defaults to throwing a [ArgumentError].
   ///
-  /// To get `null` when [path] does not point to a value in the [YamlNode]-tree,
-  /// simply pass `orElse: () => null`.
+  /// To get a default value when [path] does not point to a value in the
+  /// [YamlNode]-tree, simply pass `orElse: () => [YamlNode]`.
```
jonasfj, line 161:
```suggestion
  /// [YamlNode]-tree, simply pass `orElse: () => ...`.
```

nit: source links don't work in code-snippets (or so I think)..

yaml_edit/lib/src/editor.dart
```diff
@@ -266,7 +264,7 @@ class YamlEditor {
 
     if (parentNode is YamlList) {
       final expected = wrapAsYamlNode(
-        [...parentNode.nodes]..[keyOrIndex] = valueNode,
+        [...parentNode.nodes]..[keyOrIndex as int] = valueNode,
```
jonasfj, line 267: What if this cast fails?

We probably have to do `isValidIndex(keyOrIndex, parentNode.length)`, and if this fails, then we have to do `_pathErrorOrElse` the same way `_traverse` would do..

You actually found a bug here, I think...

```yaml
['a','b']
```

If I was to do `editor.update([3], 'c')` it'll probably fail with some unfortunate error, rather than a `PathError` (which is extends ArgumentError).

Same if I were to do: `editor.update(["not-a-number"], 'c')`

yaml_edit/lib/src/editor.dart
```diff
@@ -438,7 +425,7 @@ class YamlEditor {
     final parentNode = _traverse(collectionPath);
 
     if (parentNode is YamlList) {
-      edit = removeInList(this, parentNode, keyOrIndex);
+      edit = removeInList(this, parentNode, keyOrIndex as int);
```
jonasfj, line 428: We should probably also check that `keyOrIndex` is actually a valid index and throw `PathError` if not from here too..

yaml_edit/lib/src/editor.dart
```diff
@@ -449,7 +436,7 @@ class YamlEditor {
           updatedYamlMap(parentNode, (nodes) => nodes.remove(keyOrIndex));
     }
 
-    _performEdit(edit, collectionPath, expectedNode);
+    _performEdit(edit ?? SourceEdit(0, 0, ''), collectionPath, expectedNode);
```
jonasfj, line 439: `edit` cannot be `null` here can it?

we can simply add an `else` handler here..
Because if `parentNode` is not a `YamlList` or `YamlMap`, then we have a case like:
```yaml
key:
  subkey: my-string
```
If I was to do `editor.remove(['key', 'subkey', 'subsubkey']);` then we should throw a `PathError` because `subkey` is a string, and we cannot remove the `subsubkey` from a string...

-----

We probably need a test case for this too.

yaml_edit/lib/src/editor.dart
```diff
@@ -572,10 +551,7 @@ class YamlEditor {
   /// tree is equal to our expectations by deep equality of values. Throws an
   /// [AssertionError] if the two trees do not match.
   void _performEdit(
-      SourceEdit edit, Iterable<Object> path, YamlNode expectedNode) {
-    ArgumentError.checkNotNull(edit, 'edit');
-    ArgumentError.checkNotNull(path, 'path');
-
+      SourceEdit edit, Iterable<Object?> path, YamlNode? expectedNode) {
```
jonasfj, line 554: Can `expectedNode` be `null`? Shouldn't callers be giving a `YamlNode` where `.value == null` ?
(I'm just asking, but it looks like that it's only `remove(` that might have to do this, right?

yaml_edit/lib/src/editor.dart
```diff
@@ -613,10 +589,8 @@ class YamlEditor {
   /// the whole tree.
   ///
   /// [SourceSpan]s in this new tree are not guaranteed to be accurate.
-  YamlNode _deepModify(YamlNode tree, Iterable<Object> path,
-      Iterable<Object> subPath, YamlNode expectedNode) {
-    ArgumentError.checkNotNull(path, 'path');
-    ArgumentError.checkNotNull(tree, 'tree');
+  YamlNode? _deepModify(YamlNode tree, Iterable<Object?> path,
+      Iterable<Object?> subPath, YamlNode? expectedNode) {
```
jonasfj, line 593:
```suggestion
      Iterable<Object?> subPath, YamlNode expectedNode) {
```

Shouldn't callers give a `YamlNode` with `.value == null` ?

yaml_edit/lib/src/editor.dart
```diff
@@ -628,11 +602,11 @@ class YamlEditor {
         throw PathError(path, subPath, tree);
       }
 
-      return wrapAsYamlNode([...tree.nodes]..[keyOrIndex] = _deepModify(
+      return wrapAsYamlNode([...tree.nodes]..[keyOrIndex as int] = _deepModify(
```
jonasfj, line 605: See here we have:
```
if (!isValidIndex(keyOrIndex, tree.length)) {
  throw PathError(path, subPath, tree);
}
```
So this cast is known to be safe, and we will throw the right error :D

yaml_edit/lib/src/editor.dart
```diff
@@ -613,10 +589,8 @@ class YamlEditor {
   /// the whole tree.
   ///
   /// [SourceSpan]s in this new tree are not guaranteed to be accurate.
-  YamlNode _deepModify(YamlNode tree, Iterable<Object> path,
-      Iterable<Object> subPath, YamlNode expectedNode) {
-    ArgumentError.checkNotNull(path, 'path');
-    ArgumentError.checkNotNull(tree, 'tree');
+  YamlNode? _deepModify(YamlNode tree, Iterable<Object?> path,
```
jonasfj, line 592:
```suggestion
  YamlNode _deepModify(YamlNode tree, Iterable<Object?> path,
```
Can this return `null`, shouldn't it just return a `YamlNode` with `.value == null` ?

yaml_edit/lib/src/map_mutations.dart
```diff
@@ -47,19 +44,16 @@ SourceEdit updateInMap(
 
 /// Performs the string operation on [yaml] to achieve the effect of removing
 /// the element at [key] when re-parsed.
-SourceEdit removeInMap(YamlEditor yamlEdit, YamlMap map, Object key) {
-  ArgumentError.checkNotNull(yamlEdit, 'yamlEdit');
-  ArgumentError.checkNotNull(map, 'map');
-
+SourceEdit? removeInMap(YamlEditor yamlEdit, YamlMap map, Object? key) {
```
jonasfj, line 47: How can this return `null` ?

yaml_edit/lib/src/map_mutations.dart
```diff
@@ -47,19 +44,16 @@ SourceEdit updateInMap(
...
   if (!containsKey(map, key)) return null;
 
   final keyNode = getKeyNode(map, key);
   final valueNode = map.nodes[keyNode];
```
jonasfj, line 51:
```suggestion
  final valueNode = map.nodes[keyNode]!;
```

This way it's closer to the code that ensures this will never fail..

yaml_edit/lib/src/utils.dart
```diff
@@ -78,9 +74,7 @@ bool isBlockNode(YamlNode node) {
 
 /// Returns the content sensitive ending offset of [yamlNode] (i.e. where the last
 /// meaningful content happens)
-int getContentSensitiveEnd(YamlNode yamlNode) {
-  ArgumentError.checkNotNull(yamlNode, 'yamlNode');
-
+int getContentSensitiveEnd(YamlNode? yamlNode) {
```
jonasfj, line 77: When can `yamlNode` be `null`? I'm guessing it can't, and we should push the null check into all callers.

yaml_edit/lib/src/editor.dart
asartalo: The index out of range is handled by `updateInList()` which throws a `RangeError`. I tried using `isValidIndex()` to check the index:

```dart
!isValidIndex(keyOrIndex, parentNode.length)
```
... but that failed tests that specifically expect a `RangeError` for outside ranges. For now I plan to use the following but let me know if you want to consolidate this so it will conform to how the other methods that use `_traverse()` handle this. 

```dart
    if (parentNode is YamlList) {
      if (keyOrIndex is! int) {
        throw PathError.unexpected(path,
            'Attempted to use "$keyOrIndex" as index on the list $parentNode');
      }
      final expected = wrapAsYamlNode(
        [...parentNode.nodes]..[keyOrIndex] = valueNode,
      );

      return _performEdit(updateInList(this, parentNode, keyOrIndex, valueNode),
          collectionPath, expected);
    }
```

yaml_edit/lib/src/editor.dart
asartalo: I'll be adding tests that specifically covers index problems with lists and found on my local copy that this is already covered because `remove()` uses `_traverse()`.

yaml_edit/lib/src/editor.dart
asartalo: Sorry this was me being unsure how `edit` can be guaranteed to have a value. I could have done:

```dart
    _performEdit(edit!, collectionPath, expectedNode);
```
...and no tests would fail. But that didn't seem right. You're probably correct though that this will always be instantiated.

What should be the value of `edit` for the else case?

About the test scenario, that's covered under `test/remove_test.dart` on line 35 "PathError if collectionPath points to a scalar".

yaml_edit/lib/src/editor.dart
asartalo: Yes and it's for that reason that I changed the signature of `_performEdit()` and `_deepModify()` to accept nullable `expectedNode`. This is similar to the issue I had with `edit` value where I wasn't sure what the default value should be.

When I changed it to:

```dart
  YamlNode remove(Iterable<Object?> path) {
    SourceEdit? edit;
    var expectedNode = wrapAsYamlNode(null);
    final nodeToRemove = _traverse(path, checkAlias: true);

    if (path.isEmpty) {
...
```
...and changed the the rest of the methods to expect a non-nullable `expectedNode`, there were no errors when running tests. However, I'm still unsure what `expectedNode` does in this context so I don't know if this is the correct approach.

yaml_edit/lib/src/editor.dart
asartalo: Will include this in the next set of changes.

yaml_edit/lib/src/editor.dart
asartalo: Yup. Also please see my comment on this behavior on `update()`.

yaml_edit/lib/src/map_mutations.dart
asartalo: This is the body for `removeInMap()`:

```dart
SourceEdit? removeInMap(YamlEditor yamlEdit, YamlMap map, Object? key) {
  if (!containsKey(map, key)) return null;
...
```
...so I had to change the return value to nullable on the 2nd pass of changes I did without checking how it would be used. After changing the test to force a fail for this scenario, it would seem that the `containsKey()` call is no longer necessary since it's covered by `_traverse()`;

```sh
00:01 +1 -1: throws PathError if collectionPath is invalid [E]                                                                                                          
  Invalid path: [d]. Failed to traverse to subpath (d)! Parent {a: 1, b: 2, c: 3} does not contain key or index d.
  package:yaml_edit/src/editor.dart 502:25  YamlEditor._pathErrorOrElse
  package:yaml_edit/src/editor.dart 479:18  YamlEditor._traverse
  package:yaml_edit/src/editor.dart 413:26  YamlEditor.remove
  test/remove_test.dart 39:11               main.<fn>.<fn>
  
00:01 +1 -2: throws PathError if collectionPath is invalid in nested path [E]                                                                                           
  Invalid path: [a, c]. Failed to traverse to subpath (a, c)! Parent {b: foo} does not contain key or index c.
  package:yaml_edit/src/editor.dart 502:25  YamlEditor._pathErrorOrElse
  package:yaml_edit/src/editor.dart 479:18  YamlEditor._traverse
  package:yaml_edit/src/editor.dart 413:26  YamlEditor.remove
  test/remove_test.dart 49:11               main.<fn>.<fn>
```

asartalo: `removeInMap()` is only used in `remove()`.

yaml_edit/lib/src/utils.dart
asartalo: Yeah, I don't remember why I did this. :(

asartalo: Now I think I remember. On `_replaceInBlockMap()` was a call to `getContentSensitiveEnd(map.nodes[key]);`. I've since pushed the null assertion there.

yaml_edit/lib/src/editor.dart
```diff
@@ -120,15 +120,12 @@ class YamlEditor {
 
   factory YamlEditor(String yaml) => YamlEditor._(yaml);
 
-  YamlEditor._(this._yaml) {
-    ArgumentError.checkNotNull(_yaml);
+  YamlEditor._(this._yaml) : _contents = loadYamlNode(_yaml) {
```
asartalo, line 123: @jonasfj I moved the `_contents` instantiation here. Is this what you meant?

yaml_edit/lib/src/editor.dart
```diff
@@ -417,17 +407,15 @@ class YamlEditor {
...
   /// '''
   /// ```
-  YamlNode remove(Iterable<Object> path) {
-    ArgumentError.checkNotNull(path, 'path');
-
-    SourceEdit edit;
-    YamlNode expectedNode;
+  YamlNode remove(Iterable<Object?> path) {
+    SourceEdit? edit;
+    var expectedNode = wrapAsYamlNode(null);
```
asartalo, line 412: Setting a default `expectedNode` here. I don't know if this is the right way to do it.

yaml_edit/lib/src/editor.dart
```diff
@@ -449,7 +437,7 @@ class YamlEditor {
           updatedYamlMap(parentNode, (nodes) => nodes.remove(keyOrIndex));
     }
 
-    _performEdit(edit, collectionPath, expectedNode);
+    _performEdit(edit!, collectionPath, expectedNode);
```
asartalo, line 440: Still not sure if this is the way to go but as you said, this never seems to be null.

yaml_edit/lib/src/editor.dart
```diff
@@ -613,10 +590,8 @@ class YamlEditor {
   /// the whole tree.
   ///
   /// [SourceSpan]s in this new tree are not guaranteed to be accurate.
-  YamlNode _deepModify(YamlNode tree, Iterable<Object> path,
-      Iterable<Object> subPath, YamlNode expectedNode) {
-    ArgumentError.checkNotNull(path, 'path');
-    ArgumentError.checkNotNull(tree, 'tree');
+  YamlNode _deepModify(YamlNode tree, Iterable<Object?> path,
```
asartalo, line 593: Changed return value to non-nullable.

yaml_edit/lib/src/map_mutations.dart
```diff
@@ -47,14 +44,9 @@ SourceEdit updateInMap(
 
 /// Performs the string operation on [yaml] to achieve the effect of removing
 /// the element at [key] when re-parsed.
-SourceEdit removeInMap(YamlEditor yamlEdit, YamlMap map, Object key) {
-  ArgumentError.checkNotNull(yamlEdit, 'yamlEdit');
-  ArgumentError.checkNotNull(map, 'map');
-
-  if (!containsKey(map, key)) return null;
```
asartalo, line 54: `removeInMap()` is only used in `remove()` where the key is guaranteed. I've removed the `containsKey()` call so the return value can be non-nullable.

yaml_edit/test/remove_test.dart
```diff
@@ -39,13 +39,46 @@ c: 3
...
 
+    test('PathError if collectionPath is invalid in nested path', () {
+      final doc = YamlEditor('''
+a:
+  b: 'foo'
+''');
+
+      expect(() => doc.remove(['d']), throwsPathError);
+    });
+
```
asartalo, line 50: Test covering invalid path.

yaml_edit/test/remove_test.dart
```diff
@@ -39,13 +39,46 @@ c: 3
...
+      final doc = YamlEditor('{ a: { b: foo } }');
+      final node = doc.remove(['a', 'b']);
+      expect(node.value, equals('foo'));
+    });
+
+    test('returns null-value node when doc is empty and path is empty', () {
+      final doc = YamlEditor('');
+      final node = doc.remove([]);
+      expect(node.value, equals(null));
+    });
```
asartalo, line 81: Test covering possible invalid indeces.

yaml_edit/test/update_test.dart
```diff
@@ -34,6 +34,11 @@ void main() {
       final doc = YamlEditor("- YAML Ain't Markup Language");
       expect(() => doc.update([0, 'a'], 'a'), throwsPathError);
     });
+
+    test('PathError in list if using a non-integer as index', () {
+      final doc = YamlEditor("{ a: ['b', 'c'] }");
+      expect(() => doc.update(['a', 'b'], 'x'), throwsPathError);
+    });
```
asartalo, line 41: Test covering invalid index for lists on `update()`.

leoafarias:
Hi guys I see that seems this issue has not moved forward. This is something I need and I am willing to help. I have close to completed a null-safety migration, but looking through the comments there have been some different decisions made. Please let me know if I can help get this through the finish line.

jonasfj:
Sorry, I'm just a bit behind on reviews..

yaml_edit/lib/src/editor.dart
jonasfj: So when will this return `null`? If the document is empty and the `path` is empty, then it doesn't return `null` does it?

It returns the `YamlNode` that has been removed. So unless the path is invalid there is an `YamlNode` that the path points to, and that node should be removed and returned.

yaml_edit/lib/src/editor.dart
jonasfj: This is fine... but this is only necessary when in the `if (path.isEmpty) {` case...

So we could make `late expectedNode` and only do `expectedNode = wrapAsYamlNode(null);` in the case where `path.isEmpty`.

yaml_edit/lib/src/editor.dart
```diff
@@ -417,17 +407,15 @@ class YamlEditor {
   /// - 2 # comment 2
   /// '''
   /// ```
-  YamlNode remove(Iterable<Object> path) {
-    ArgumentError.checkNotNull(path, 'path');
-
-    SourceEdit edit;
-    YamlNode expectedNode;
+  YamlNode remove(Iterable<Object?> path) {
+    SourceEdit? edit;
```
jonasfj, line 411: How about:
```suggestion
    late SourceEdit edit;
```

All the if-branches below will assign a value to `edit`. So we will write to it before we read from it, no matter what happens.

yaml_edit/lib/src/editor.dart
jonasfj: I would suggest making `edit` a `late SourceEdit edit`, but that's a nit :D

yaml_edit/lib/src/editor.dart
jonasfj: Love it!

yaml_edit/lib/src/editor.dart
jonasfj: You can also use `late SourceEdit;`

yaml_edit/lib/src/map_mutations.dart
jonasfj: We could add an `assert(containsKey(map, key))` for extra sanity if we wanted to...

jonasfj (approved):
I think the comments from my last review are just internal nits... @asartalo I suggest addressing them in a follow up PR..

I'll make a version 2.0.0, update changelog and publish.. sorry for the turn around.

jonasfj:
@leoafarias, published as 2.0.0... curious what are you using it for?

@asartalo, please feel free to file PRs addressing the nits pointed out, I don't think they are important, just nice..
You are also always welcome to file more PRs with additional tests and documentation examples :D

leoafarias:
@jonasfj https://github.com/leoafarias/pkg

I have to modify the pubspec, but also want to have something that has minimal dependencies. This package was a life saver, and the only one I found ;)

Thanks for prioritizing this!

asartalo:
@jonasj Sorry. Life got in the way that I forgot all about this. I'll take a look at the things that you raised but it might take me a while to get up to them.

yaml_edit/lib/src/editor.dart
asartalo: The signature has been changed and this no longer applies. It now returns just `YamlNode`.

PR #110, opened 2021-03-14
[yaml_edit] Testcases for `parseAt()` with null values

asartalo:
This is just additional coverage for parsing YAML documents with null values.

@jonasfj

### DIFF
--- a/yaml_edit/test/parse_test.dart
+++ b/yaml_edit/test/parse_test.dart
@@ -100,13 +100,22 @@ void main() {
     test('with the correct value in nested collection', () {
       final doc = YamlEditor('''
 a: 1
-b: 
+b:
   d: 4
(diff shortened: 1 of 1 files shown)


jonasfj (approved):

jonasfj:
@asartalo any reason we shouldn't just land this as is?

We're just testing current behavior, and it's unlikely that we want to change this behavior due to null-safety migration, is it not?

asartalo:
> @asartalo any reason we shouldn't just land this as is?
> 
> We're just testing current behavior, and it's unlikely that we want to change this behavior due to null-safety migration, is it not?

Yes, I think this would be good to merge. I was using this to prove the null parsing behavior on master.

jonasfj (approved):

PR #125, opened 2021-04-26
Updated dependencies

leoafarias:
Packages that are dependent on this are not being able to migrate to null-safety due to the `yaml` dep being out of date.


leoafarias:
This is not needed due to #107

PR #137, opened 2021-06-18
Null-safety migration of neat_cache

jonasfj:
Tests not passing yet :D

### DIFF
--- a/neat_cache/lib/neat_cache.dart
+++ b/neat_cache/lib/neat_cache.dart
@@ -102,7 +102,7 @@ abstract class Entry<T> {
   /// The [get] method is a best-effort method. In case of intermittent failures
   /// from the underlying [CacheProvider] the [get] method will ignore failures
   /// and return `null` (or result from [create] if specified).
-  Future<T> get([Future<T> Function() create, Duration ttl]);
+  Future<T?> get([Future<T?> Function() create, Duration ttl]);
 
   /// Set the value stored in this cache entry.
   ///
@@ -116,7 +116,7 @@ abstract class Entry<T> {
   ///
   /// To ensure that cache entries are purged, use the [purge] method with
   /// `retries` not set to zero.
-  Future<T> set(T value, [Duration ttl]);
+  Future<T?> set(T? value, [Duration ttl]);
 
   /// Clear the value stored in this cache entry.
   ///
@@ -131,7 +131,7 @@ class _Cache<T, V> implements Cache<T> {
   final CacheProvider<V> _provider;
   final String _prefix;
   final Codec<T, V> _codec;
-  final Duration _ttl;
+  final Duration? _ttl;
 
   _Cache(this._provider, this._prefix, this._codec, [this._ttl]);
 
@@ -156,8 +156,8 @@ class _Entry<T, V> implements Entry<T> {
   _Entry(this._owner, this._key);
 
   @override
-  Future<T> get([Future<T> Function() create, Duration ttl]) async {
-    V value;
+  Future<T?> get([Future<T?> Function()? create, Duration? ttl]) async {
+    V? value;
     try {
       _logger.finest(() => 'reading cache entry for "$_key"');
       value = await _owner._provider.get(_key);
@@ -175,7 +175,11 @@ class _Entry<T, V> implements Entry<T> {
       final created = await create();
       if (created != null) {
         // Calling `set(null)` is equivalent to `purge()`, we can skip that here
-        await set(created, ttl);
+        if (ttl != null) {
+          await set(created, ttl);
+        } else {
+          await set(created);
+        }
       }
       return created;
     }
@@ -183,7 +187,7 @@ class _Entry<T, V> implements Entry<T> {
   }
 
   @override
-  Future<T> set(T value, [Duration ttl]) async {
+  Future<T?> set(T? value, [Duration? ttl]) async {
     if (value == null) {
       await purge();
       return null;
@@ -191,7 +195,11 @@ class _Entry<T, V> implements Entry<T> {
     ttl ??= _owner._ttl;
     final raw = _owner._codec.encode(value);
     try {
-      await _owner._provider.set(_key, raw, ttl);
+      if (ttl != null) {
+        await _owner._provider.set(_key, raw, ttl);
+      } else {
+        await _owner._provider.set(_key, raw);
+      }
     } on IntermittentCacheException {
       _logger.fine(
         // embedding [intermittent-cache-failure] to allow for easy log metrics
--- a/neat_cache/lib/neat_cache.dart
+++ b/neat_cache/lib/neat_cache.dart
@@ -175,11 +175,7 @@ class _Entry<T, V> implements Entry<T> {
       final created = await create();
       if (created != null) {
         // Calling `set(null)` is equivalent to `purge()`, we can skip that here
-        if (ttl != null) {
-          await set(created, ttl);
-        } else {
-          await set(created);
-        }
+        await set(created, ttl);
       }
       return created;
     }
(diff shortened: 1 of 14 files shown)


neat_cache/lib/neat_cache.dart
```diff
@@ -175,23 +175,31 @@ class _Entry<T, V> implements Entry<T> {
       final created = await create();
       if (created != null) {
         // Calling `set(null)` is equivalent to `purge()`, we can skip that here
-        await set(created, ttl);
+        if (ttl != null) {
+          await set(created, ttl);
+        } else {
+          await set(created);
```
isoos, line 181: why can't we use `set(created, ttl)` unconditionally?

neat_cache/lib/neat_cache.dart
```diff
@@ -175,23 +175,31 @@ class _Entry<T, V> implements Entry<T> {
...
       return null;
     }
     ttl ??= _owner._ttl;
     final raw = _owner._codec.encode(value);
     try {
-      await _owner._provider.set(_key, raw, ttl);
+      if (ttl != null) {
+        await _owner._provider.set(_key, raw, ttl);
+      } else {
+        await _owner._provider.set(_key, raw);
```
isoos, line 201: why can't we use `set(_key, raw, ttl)` unconditionally?

neat_cache/lib/src/providers/redis.dart
```diff
@@ -54,19 +53,22 @@ class RedisCacheProvider extends CacheProvider<List<int>> {
...
-        commands: client.asCommands(),
+        client: RespClient(socket, socket),
       );
+    } on RedisConnectionException {
+      throw IntermittentCacheException('connection failed');
     } on TimeoutException {
       throw IntermittentCacheException('connect failed with timeout');
-    } on SocketException {
+    } on IOException {
       throw IntermittentCacheException('connect failed with socket exception');
```
isoos, line 72: Also update the exception text, eg: `'connect failed with IO exception'`.

isoos, line 72: we should probably also add the exception message here.

neat_cache/lib/neat_cache.dart
jonasfj: Because when you set a `ttl` you can'tse it to `null`, you can choose not to set it, but not set it to `null`.

We can debate the wisdom of such design... I'm okay changing it :D

isoos: I don't see much harm in that, especially in cases like this, where there could be many different fallback config parameter that the client will use in the presence of `null`.

neat_cache/lib/src/providers/redis.dart
jonasfj: Adding... `$e`, but in general, we just ignore these exceptions... they are signals to try-again :D

neat_cache/lib/neat_cache.dart
jonasfj: If I allow `Duration?` in the `CacheProvider` abstraction... then we don't need the conditional here... and then I think we'll all be reasonably happy :D

neat_cache/lib/src/providers/inmemory.dart
```diff
@@ -104,7 +104,7 @@ class InMemoryCacheProvider<T> extends CacheProvider<T> {
   @override
   Future<void> close() async {
     _isClosed = true;
-    _old = null;
-    _new = null;
+    _old = {};
```
isoos, line 107: nit: if the `_old` and `_new` fields are final, this could be `_old.clear()` instead. not sure if it makes any difference...

neat_cache/lib/src/providers/resp.dart
```diff
@@ -0,0 +1,514 @@
...
+import 'dart:collection';
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:logging/logging.dart';
+
+final _log = Logger('neat_cache:redis');
+
+/// Thrown when a command returns
+class RedisErrorException implements Exception {
```
isoos, line 30: This name is a bit strange, why not just `RedisException` or maybe `RedisServerException`?

neat_cache/lib/src/providers/resp.dart
```diff
@@ -0,0 +1,514 @@
...
+import 'dart:async';
+import 'dart:collection';
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:logging/logging.dart';
+
+final _log = Logger('neat_cache:redis');
+
+/// Thrown when a command returns
```
isoos, line 29: Maybe `Thrown when a command returns with an error.`?

neat_cache/lib/src/providers/resp.dart
```diff
@@ -0,0 +1,514 @@
...
+  String toString() => 'RedisErrorException: $type $message';
+}
+
+/// Thrown if the redis connection is broken.
+///
+/// This typically happens if the connection is unexpectedly closed, the
+/// [RESP protocol][1] is violated, or there is an internal error.
+///
+/// [1]: https://redis.io/topics/protocol
+class RedisConnectionException implements Exception {
```
isoos, line 46: Could this be an `IOException`? or a `FormatException`?

neat_cache/lib/src/providers/resp.dart
```diff
@@ -0,0 +1,514 @@
...
+        }
+        if (_pending.isEmpty) {
+          return await _abort(
+            RedisConnectionException._('unexpected value from server'),
+            StackTrace.current,
+          );
+        }
+        final c = _pending.removeFirst();
+        if (value is RedisErrorException) {
+          // This is an error code returned by the server, it doesn't have a
```
isoos, line 268: Wouldn't it be still useful if we'd keep the caller's stacktrace context?

neat_cache/lib/src/providers/resp.dart
```diff
@@ -0,0 +1,514 @@
...
+      }
+    }
+
+    // Handle errors
+    if (type == '-'.codeUnitAt(0)) {
+      final message = utf8.decode(
+        rest,
+        allowMalformed: true,
+      );
+      final i = message.indexOf(RegExp(r'\s'));
```
isoos, line 324: Let's extract this into a top-level variable.

neat_cache/lib/src/providers/resp.dart
jonasfj: hmm, maybe `RedisCommandException` is better, because it due to an error from a command usually...

isoos: sgtm

neat_cache/lib/src/providers/inmemory.dart
jonasfj: If they are final you can't swap them...

neat_cache/lib/src/providers/resp.dart
jonasfj: Are you suggesting reuse those instead, or subclassing them?

I found that there is a lot of exception that a consumer of `resp.dart` doesn't need to know about... The essence is the same: connection has stopped working and you need to reconnect. That's what this exception is.

neat_cache/lib/src/providers/resp.dart
jonasfj: Maybe, but won't that get appended when you await the future?

neat_cache/lib/src/providers/resp.dart
isoos: I was thinking `implements IOException`. I think it is fine like this, but for database-like clients I like to have the separation of the errors being IO and non-IO...

neat_cache/lib/src/providers/resp.dart
isoos:
```dart
import 'dart:async';

Future<void> main() async {
  await _method();
}

Future<void> _method() async {
  final c = Completer();
  final f = c.future;
  c.completeError('', StackTrace.empty);
  await f;
}
```
It won't list neither the `main` nor the `_method` in the output. If you change to StackTrace.current, it will do both.

neat_cache/lib/src/providers/resp.dart
jonasfj: I just have mixed feelings about dependency on `dart:io`.

isoos (approved):

neat_cache/lib/src/providers/resp.dart
```diff
@@ -182,7 +182,15 @@ class RespClient {
       await _abort(e, st);
     }
 
-    return c.future;
+    try {
+      return await c.future;
+    } on RedisCommandException catch (e) {
```
isoos, line 187: nice!

Issue #136, opened 2021-06-13
Unable to mount multiple handlers to the same prefix


olefasting:
This was originally posted as an issue with shelf, but I guess this is the right place for this:

Say that I have two services (an authentication service and a profiles service, in my case) and an API service that is meant to bind it all together into a final handler. If I want to prefix the two services with an /api' prefix, by mounting both to /api/ in the API service, only the first handler will be accessible when running the server.

I have tested that everything works, otherwise, by mounting the two services to different prefixes, so the issue seems to lie in the fact that the mounts are both to the same prefix...

This seems like a bug to me, as it is normal practice in libraries like express, which shelf seems to emulate...

Also, when I remap the authentication service to / and keep the profiles service at /api/, the latter will only work if it is mapped before the former. Very strange behavior, if you ask me....

```dart
class AuthenticationService {
  AuthenticationService();

  Handler get handler {
    final router = Router();

    /* Various handlers mapped to /login, /register etc */

    return router;
  }
}

class ProfileService {
  ProfileService();

  Handler get handler {
    final router = Router();

    /* Various handlers, all mapped to /profile, but with different methods */

    return router;
  }
}

class ApiService {
  Handler get handler {
    final router = Router();

    final authenticationService = AuthenticationService();
    final profileService = ProfileService();

    router.mount('/api/', profileService.handler); // Only this will work. If I change the order, it is always the first mount that works....
    router.mount('/api/', authenticationService.handler);

    return router;
  }
}
```

jonasfj:
I think we did consider implementing the semantics from [Cascade](https://pub.dev/documentation/shelf/latest/shelf/Cascade-class.html), but then it wouldn't be possible to return a custom 404 for a custom route, and have a different catch-all 404 for all routes.

For now consider using [Cascade](https://pub.dev/documentation/shelf/latest/shelf/Cascade-class.html) for mounts under `api/`.

I do see the problem, but this would be a breaking change, please have people vote up if they think this is desired behavior.

jonasfj:
Let's introduce a sentinel `final routeNotFound = shelf.Response.notFound('route not found');`, and whenever a router has no matching route it returns the `routeNotFound` sentinel object.

Whenver a router gets a response from a handler or sub-router it will check if response is the sentinel (`response == routeNotFound`) if this is the case then the response is ignored and other routes are matched.

To do:
 *  [ ] Add sentinel
 *  [ ] Update documentation for router and example
 *  [ ] Add tests
[...]

PR #140, opened 2021-06-30
[DRAFT] Remove constraint to mount routes without trailing slash (fixes #67)

### LINKED ISSUE #67: Mounts should not require trailing slashes
* When using the mount function a trailing slash is required.
* This makes it impossible to have a mounted route that does not require that a tailing slash be required
* In example below the URL "/product" would result in a 404
* "/product/" would return the expected result

```dart
void main(List<String> args) async {
  load();

  var router = Router();

  router.mount('/auth/', AuthRoutes().router);
  router.mount('/product/', ProductRoutes().router);

  // Final route to return a not found
  router.all('/<.*>', (request) => shelf.Response.notFound(''));

  var handler = const shelf.Pipeline()
    .addMiddleware(shelf.logRequests())
    .addMiddleware(shelf.createMiddleware(requestHandler: AuthManager.handle))
    .addMiddleware(shelf.createMiddleware(requestHandler: ProductController.SecureMiddleware))
    .addHandler(router.handler);

  await io.serve(handler, 'localhost', int.parse(env['PORT'])).then((server) {
    print('Server running on port ${server.port}');
  });

}

class ProductRoutes {

  static FutureOr<Response> middleware(Request request) async {
    print("ProductRoutes Middleware");
    return null;
  }

  Router get router {
    final router = Router();

    var c = ProductController();

    router.get('/', c.GetList);
    router.get('/<product_id>', c.GetById);

    return router;
  }

}
```

matsp:
Hi. 

This MR is still a draft, because the added testcase is still failing. I am not sure if this a new bug or has something to do with the code change. Till now I was not able to figure out the root case for it.

The failing testcase tries to mount a router (has prefix '/') on another handler with the prefix '/api2'. There is also another handler mounted at '/api' which get called instead of '/api2'. If you change the order it works. 

I think this has something to do with the pattern matching or the creation of the regex pattern at the RouteEntry class. For '/api2' the params 'path: 2' will be generated which can't be correct.

### DIFF
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -120,7 +120,7 @@ class Router {
   ///
   /// In this case prefix may not contain any parameters, nor
   void mount(String prefix, Handler handler) {
-    if (!prefix.startsWith('/') || !prefix.endsWith('/')) {
+    if (!prefix.startsWith('/')) {
       throw ArgumentError.value(
           prefix, 'prefix', 'must start and end with a slash');
     }
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -121,8 +121,7 @@ class Router {
   /// In this case prefix may not contain any parameters, nor
   void mount(String prefix, Handler handler) {
     if (!prefix.startsWith('/')) {
-      throw ArgumentError.value(
-          prefix, 'prefix', 'must start and end with a slash');
+      throw ArgumentError.value(prefix, 'prefix', 'must start with a slash');
     }
 
     // first slash is always in request.handlerPath
--- a/shelf_router/test/router_test.dart
+++ b/shelf_router/test/router_test.dart
@@ -103,12 +103,18 @@ void main() {
       return Response.ok('Hello $user');
     });
 
+    var api2 = Router();
+    api2.get('/', (Request request) {
+      return Response.ok('api2');
+    });
+
     var app = Router();
     app.get('/hello', (Request request) {
       return Response.ok('hello-world');
     });
 
-    app.mount('/api/', api);
+    app.mount('/api', api);
+    app.mount('/api2', api2);
 
     app.all('/<_|[^]*>', (Request request) {
       return Response.notFound('catch-all-handler');
@@ -118,6 +124,8 @@ void main() {
 
     expect(await get('/hello'), 'hello-world');
     expect(await get('/api/user/jonasfj/info'), 'Hello jonasfj');
+    expect(await get('/api2'), 'api2');
+    expect(get('/api2/'), throwsA(anything));
     expect(get('/api/user/jonasfj/info-wrong'), throwsA(anything));
   });
 
@@ -138,7 +146,7 @@ void main() {
 
     var app = Router();
     app.mount(
-      '/api/',
+      '/api',
       Pipeline().addMiddleware(middleware).addHandler(api),
     );
 


jonasfj:
I think this is a similar to https://github.com/google/dart-neats/issues/136.

Maybe we should consider changing behavior again, such that 404s are always ignored. That makes it kind of impossible to make custom 404s for a given route.

Another option is that we a special 404 instance, so that only that instance will be ignored and tried again.

isoos:
> Maybe we should consider changing behavior again, such that 404s are always ignored. That makes it kind of impossible to make custom 404s for a given route.
> Another option is that we a special 404 instance, so that only that instance will be ignored and tried again.

This was also an issue in pub.dev's migration. I think a custom, easy-to-compare 404 instance, e.g. along the lines of "Route not found." would be great to have.

matsp:
@jonasfj Is there a task that could be derived? I think this behavior should really be changed to make it most transparent to the users.

jonasfj:
I think this is fixed with `shelf_router` version `1.1.0`, @matsp let me know if this is still a problem.

I think you issue had to do with nested-routers, for which `1.1.0` has changed behavior, see `CHANGELOG`.

jonasfj:
hmm. @matsp, I think 1.1.0 only solved the problem for nested-routers.. not necessarily that case you point out in https://github.com/google/dart-neats/issues/67, my bad, this PR is likely still relevant.

matsp:
@jonasfj It still looks like that the 'api2' call is directed to the 'api' handler. I was not able to find the root cause yet.

shelf_router/lib/src/router.dart
```diff
@@ -147,9 +147,8 @@ class Router {
...
   /// In this case prefix may not contain any parameters, nor
   void mount(String prefix, Handler handler) {
-    if (!prefix.startsWith('/') || !prefix.endsWith('/')) {
-      throw ArgumentError.value(
-          prefix, 'prefix', 'must start and end with a slash');
+    if (!prefix.startsWith('/')) {
+      throw ArgumentError.value(prefix, 'prefix', 'must start with a slash');
     }
 
     // first slash is always in request.handlerPath
```
jonasfj, line 154: Hmm, maybe we need to do:

```dart
    final path = prefix.substring(1);
    if (prefix.endsWith('/')) {
      all(prefix + '<path|[^]*>', (Request request) {
        return handler(request.change(path: path));
      });
    } else {
      all(prefix, (Request request) {
        return handler(request.change(path: path));
      });
      all(prefix + '/<path|[^]*>', (Request request) {
        return handler(request.change(path: path + '/'));
      });
    }
```

Aren't we really trying to handle two cases here? The one with a trailing slash and the one without a trailing slash.
And if we mount without a trailing slash, then we have two patterns, `/prefix` and `/prefix/<whatever>`.

I do wonder if `request.change(path: 'prefix')` will actually work with a nested router that's trying to match `/`, but that's probably something we could make a test case for.

jonasfj: > It still looks like that the 'api2' call is directed to the 'api' handler. I was not able to find the root cause yet.

Yeah, that because when prefix is `/api` then your pattern becomes `/api<path|[^]*>` which is probably not what you want. As a request for `/api2` will match pattern `/api<path|[^]*>`.

Hence, I propose that we make two patterns as outlined above.

matsp:
closing this due #146 has more progress with the mentioned patch above

PR #141, opened 2021-07-07
Introduce `routeNotFound` sentinel for shelf_router 1.1.0

jonasfj:
Fixing https://github.com/google/dart-neats/issues/136

### DIFF
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -12,22 +12,48 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:collection' show UnmodifiableMapView;
+
 import 'package:http_methods/http_methods.dart';
 import 'package:meta/meta.dart' show sealed;
 import 'package:shelf/shelf.dart';
 import 'package:shelf_router/src/router_entry.dart' show RouterEntry;
 
 /// Get a URL parameter captured by the [Router].
+@Deprecated('Use Request.params instead')
 String params(Request request, String name) {
-  final p = request.context['shelf_router/params'];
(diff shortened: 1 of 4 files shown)


shelf_router/lib/src/router.dart
```diff
@@ -184,7 +214,58 @@ class Router {
...
+  /// // response with message 'nothing found'.
+  ///
+  /// // In the pattern below `<anything|.*>` is on the form `<name|regex>`,
+  /// // thus, this simply creates a URL parameter called `anything` which
+  /// // matches anything.
+  /// app.all('/<anything|.*>', (Request request) {
+  ///   return Response.notFound('nothing found');
+  /// });
+  /// ```
+  static final Response routeNotFound = Response.notFound('Route not found');
```
isoos, line 270: Nit: I was thinking that this will be top-level, but it may be even better inside the `Router`.

isoos (approved):

shelf_router/lib/src/router.dart
jonasfj: We could do:
A) Top-level `routeNotFound`,
B) `Router.routeNotFound`, or,
C) `Route.notFound`.

I'm not 100% sure... I think I'm leaning away from (A) because this isn't a thing most users will need to use. Mostly you only need it if mounting nested routers, and since routers return `routeNotFound` by default, you won't need to know about this. I'm hoping that for most users this is an internal technicality, that makes nested routers "just work".

(C) `Route` is an annotation for `shelf_router_generator`, so adding it there seems a bit unfortunate.

So that's why I went with (B), hide it away, because most users shouldn't think about this.

PR #142, opened 2021-07-07
Fix multiple commands in redis cache.

isoos:


neat_cache/lib/src/providers/resp.dart
```diff
@@ -177,7 +177,7 @@ class RespClient {
     final c = Completer<Object?>();
     _pending.addLast(c);
     try {
-      await _output.addStream(Stream.value(out.toBytes()));
+      _output.add(out.toBytes());
```
jonasfj, line 180: I think it might be better to use a future to ensure that `_output.addStream` is called sequentially..

The problem with not awaiting it that if sending to the server fails, the exception won't cause the `RespClient` to close, hence, it might keep accepting commands.

Technically, I hope the stream will at some point enter a state where it throw exceptions synchronously, but without knowing the underlying implementation I'm not that will happen.

jonasfj (approved):

jonasfj:
Published as `neat_cache` version `2.0.1`.

PR #145, opened 2021-07-13
fix: StateError on back-to-back routeNotFound handler invocations

### LINKED ISSUE #144: StateError (Bad state: The 'read' method can only be called once on a shelf.Request/shelf.Response object.)
* When back-to-back requests are made with the `routeNotFound` response, a `StateError` is thrown because the same `Response` object is returned multiple times and `read` can only be called once.
* Unable to reproduce with v1.0.0
  * Seems like this was introduced in v1.1.0 by including the `routeNotFound` sentinel.

felangel:
closes #144

### DIFF
--- a/shelf_router/lib/src/router.dart
+++ b/shelf_router/lib/src/router.dart
@@ -163,6 +163,14 @@ class Router {
   ///
   /// This method allows a Router instance to be a [Handler].
   Future<Response> call(Request request) async {
+    bool _isNotDefaultNotFound(Response response) {
+      return response.statusCode != routeNotFound.statusCode ||
+          response.contentLength != routeNotFound.contentLength ||
+          response.context != routeNotFound.context ||
+          response.headers != routeNotFound.headers ||
+          response.mimeType != routeNotFound.mimeType;
+    }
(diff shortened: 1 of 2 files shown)


shelf_router/lib/src/router.dart
```diff
@@ -163,6 +163,14 @@ class Router {
   ///
   /// This method allows a Router instance to be a [Handler].
   Future<Response> call(Request request) async {
+    bool _isNotDefaultNotFound(Response response) {
+      return response.statusCode != routeNotFound.statusCode ||
```
felangel, line 167: Is there a better way to determine whether the response is the default without having a static response?

isoos:
I think instead of this, we should change the `routeNotFound` to implement `shelf.Response`, and use a repeatable read method that doesn't throw in such cases.

felangel:
> I think instead of this, we should change the `routeNotFound` to implement `shelf.Response`, and use a repeatable read method that doesn't throw in such cases.

I just took a stab at implementing this. Let me know what you think and happy to adjust 👍

PR #146, opened 2021-07-14
fix: mount remove trailing slash requirement

### LINKED ISSUE #67: Mounts should not require trailing slashes

felangel:
- closes #67 (related to #140)

### DIFF
--- a/shelf_router/test/router_test.dart
+++ b/shelf_router/test/router_test.dart
@@ -148,6 +148,30 @@ void main() {
     expect(await get('/api/hello?ok'), 'middleware');
   });
 
+  test('mount(Router) does not require a trailing slash', () async {
+    var api = Router();
+    api.get('/', (Request request) {
+      return Response.ok('Hello World!');
+    });
+
+    var app = Router();
+    app.get('/hello', (Request request) {
+      return Response.ok('hello-world');
+    });
+
+    app.mount('/api', api);
+
+    app.all('/<_|[^]*>', (Request request) {
+      return Response.ok('catch-all-handler');
+    });
+
+    server.mount(app);
+
+    expect(await get('/hello'), 'hello-world');
+    expect(await get('/api'), 'Hello World!');
+    expect(await get('/api/user/jonasfj/info-wrong'), 'catch-all-handler');
+  });
+
   test('responds with 404 if no handler matches', () {
     var api = Router()..get('/hello', (request) => Response.ok('Hello'));
     server.mount(api);
--- a/shelf_router/test/router_test.dart
+++ b/shelf_router/test/router_test.dart
@@ -154,6 +154,10 @@ void main() {
       return Response.ok('Hello World!');
     });
 
+    api.get('/user/<user>/info', (Request request, String user) {
+      return Response.ok('Hello $user');
+    });
+
     var app = Router();
     app.get('/hello', (Request request) {
       return Response.ok('hello-world');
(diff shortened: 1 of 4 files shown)


shelf_router/test/router_test.dart
```diff
@@ -148,6 +148,30 @@ void main() {
     expect(await get('/api/hello?ok'), 'middleware');
   });
 
+  test('mount(Router) does not require a trailing slash', () async {
+    var api = Router();
+    api.get('/', (Request request) {
+      return Response.ok('Hello World!');
+    });
```
jonasfj, line 155: Can we make another end-point on `api`, so that we have test that this works for something like `/api/user/jonasfj/info`?

shelf_router/test/router_test.dart
```diff
@@ -148,6 +148,30 @@ void main() {
...
+    app.mount('/api', api);
+
+    app.all('/<_|[^]*>', (Request request) {
+      return Response.ok('catch-all-handler');
+    });
+
+    server.mount(app);
+
+    expect(await get('/hello'), 'hello-world');
+    expect(await get('/api'), 'Hello World!');
```
jonasfj, line 171: Both of these should work, right?
```suggestion
    expect(await get('/api'), 'Hello World!');
    expect(await get('/api/'), 'Hello World!');
```

jonasfj (changes requested):
Looks good, but it would be nice to have a few more tests -- just to be sure (also to guard against regressions in the future).

shelf_router/test/router_test.dart
felangel: yup just added the additional expectation 👍

felangel:
> Looks good, but it would be nice to have a few more tests -- just to be sure (also to guard against regressions in the future).
> 
> Also:
> 
> * [X]  Bump minor version in `pubspec.yaml`,
> * [X]  Update `CHANGELOG.md`
> * [X]  Extend tests to cover `/api`, `/api/`, `/api/user/jonasfj/info`

Sorry for the delay! I updated the test to cover the additional cases you mentioned and update the package version and `CHANGELOG`. Let me know what you think, thanks!

felangel:
@jonasfj any updates on this? Is there anything I can do to push this forward?

jonasfj:
Closed and reopen to retrigger CI.

felangel:
Bump

jonasfj (approved):

PR #154, opened 2021-08-02
Send QUIT command when closing the connection without force: true.

isoos:
As we set `_closing = true` in the line above the if condition, I think the expression should depend on the `force` parameter.

### DIFF
--- a/neat_cache/lib/src/providers/resp.dart
+++ b/neat_cache/lib/src/providers/resp.dart
@@ -203,7 +203,7 @@ class RespClient {
   Future<void> close({bool force = false}) async {
     _closing = true;
 
-    if (!_closing) {
+    if (!force) {
       // Always send QUIT message to be nice
       try {
         final quit = command(['QUIT']);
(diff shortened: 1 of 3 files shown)


neat_cache/lib/src/providers/resp.dart
```diff
@@ -203,7 +203,7 @@ class RespClient {
   Future<void> close({bool force = false}) async {
     _closing = true;
 
-    if (!_closing) {
```
sigurdm, line 206: How was this even supposed to work before?

neat_cache/lib/src/providers/resp.dart
```diff
@@ -203,7 +203,7 @@ class RespClient {
...
+    if (!force) {
```
sigurdm, line 206: Hmm, I think the dartdoc could b e improved. Perhaps something like:

```
///  Close the connection to redis and send a `QUIT` command unless `force`.
```

Otherwise we should perhaps just always send it?

neat_cache/lib/src/providers/resp.dart
isoos: I think we have never sent the `QUIT` message.

sigurdm (approved):
LGTM

lhmzhou (approved):

Issue #163, opened 2021-12-08
get current attempt times


zysc8313:
Would that be possible that I can get current attempt times in the onRetry callback function?

ibhanu:
@zysc8313 you can checkout this: [https://github.com/ibhanu/dart-neats/tree/master/retry](https://github.com/ibhanu/dart-neats/tree/master/retry)

I don't know is it PR worthy.

zysc8313:
@ibhanu Thank you! that's definitely a good PR

ibhanu:
@zysc8313 done [https://github.com/google/dart-neats/pull/164](https://github.com/google/dart-neats/pull/164)

jonasfj:
This is breaking, I'd suggest making a different better retry package :)

PR #164, opened 2021-12-29
feat: returning current number of attempt [solving for issue #163]

ibhanu:
Here

### DIFF
--- a/retry/lib/retry.dart
+++ b/retry/lib/retry.dart
@@ -120,8 +120,8 @@ class RetryOptions {
   /// as an [Exception].
   Future<T> retry<T>(
     FutureOr<T> Function() fn, {
-    FutureOr<bool> Function(Exception)? retryIf,
-    FutureOr<void> Function(Exception)? onRetry,
+    FutureOr<bool> Function(Exception, int)? retryIf,
+    FutureOr<void> Function(Exception, int)? onRetry,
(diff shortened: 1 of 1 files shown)


retry/lib/retry.dart
```diff
@@ -120,8 +120,8 @@ class RetryOptions {
   /// as an [Exception].
   Future<T> retry<T>(
     FutureOr<T> Function() fn, {
-    FutureOr<bool> Function(Exception)? retryIf,
-    FutureOr<void> Function(Exception)? onRetry,
+    FutureOr<bool> Function(Exception, int)? retryIf,
```
isoos, line 123: This is a breaking change for all the libraries that are using this method, and should be reflected in version and CHANGELOG as well. The method also needs more documentation on what the `int` means.

I'm wondering if there is a different signature that is not a breaking change, e.g. with a different callback parameter?

jonasfj: Yeah, we could do: `FutureOr<boo> Function(Exception e, int attempt) retryAt`

But it doesn't super well, and IMO, I'm not sure there is a good way to limit retries based on the kind of exception thrown.
You won't remember if previous attempts failed due to some other `Exception`.

jonasfj:
I don't think we want breaking changes in `package:retry`.

PR #167, opened 2022-02-21
orElse callback when retries are exhausted

isoos:


retry/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v3.2.0
+ * `orElse` callback to calculate a value when the attempts where exhausted.
```
jonasfj, line 2:
```suggestion
 * Introduced `orElse` callback to create a value when the retry attempts are exhausted.
```

jonasfj (changes requested):

retry/CHANGELOG.md
isoos: Hm, yeah, it may be confusing. Maybe `onRetryExhausted` could be used instead?

retry/lib/retry.dart
```diff
@@ -122,6 +122,7 @@ class RetryOptions {
     FutureOr<T> Function() fn, {
     FutureOr<bool> Function(Exception)? retryIf,
     FutureOr<void> Function(Exception)? onRetry,
+    FutureOr<T> Function(Exception)? orElse,
```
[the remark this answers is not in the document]
jonasfj: API docs is still missing :D

slovnicki:
@isoos @jonasfj How's this progressing?
I found this PR because I needed exactly this feature.

retry/lib/retry.dart
```diff
@@ -132,6 +133,10 @@ class RetryOptions {
       } on Exception catch (e) {
         if (attempt >= maxAttempts ||
             (retryIf != null && !(await retryIf(e)))) {
+          if (orElse != null) {
+            return await orElse(e);
+          }
```
jonasfj, line 138: There are two kinds of exceptions here:
 * (A) Exceptions that trigger a retry -- those where `retryIf(e)` returns true.
 * (B) Exceptions that don't trigger a retry.

There are two reasons a call to `retry(fn, ...)` may throw:
 * (i) `fn` throws an exception of type (B),
 * (ii) `fn` throws an exception of type (A) enough times that the number of retries have been exhausted.

Do we think it's right to call `orElse` in both cases (i) and (ii)?

jonasfj:
@slovnicki, feel free to make a new PR, or comment on this one. I don't think we've resolve the semantics.

side note: I don't think we want breaking changes, hence throwing a `RetryExhaustedException` is not really a good option.

retry/lib/retry.dart
isoos: Oh, this is an old one that is still open :)

I think we should only have an alternative callback for (ii), and with that in mind, `orElse` is not the best name for it. How about `afterExhausted` or `onAttemptsExhausted`?

jonasfj: I think `orElse` is an okay name, but we might have to document the behavior -- if we want this at all.

drown0315:
I think to add stackTrace parameter for the orElse method would better

PR #211, opened 2023-07-08
feat: add stacktrace parameter on retry

drown0315:
Add stacktrace parameter on retryIf and onRetry function for getting additional information


retry/lib/retry.dart
```diff
@@ -177,8 +177,8 @@ Future<T> retry<T>(
   double randomizationFactor = 0.25,
   Duration maxDelay = const Duration(seconds: 30),
   int maxAttempts = 8,
-  FutureOr<bool> Function(Exception)? retryIf,
-  FutureOr<void> Function(Exception)? onRetry,
+  FutureOr<bool> Function(Exception, StackTrace)? retryIf,
```
jonasfj, line 180: I don't think users should be allowed to condition the retry on the stacktrace.

if you're doing this something is very wrong. It's probably better to fix the thing that is wrong.

retry/lib/retry.dart
```diff
@@ -177,8 +177,8 @@ Future<T> retry<T>(
...
+  FutureOr<void> Function(Exception, StackTrace)? onRetry,
```
jonasfj, line 181: This could be useful, but is it worth doing a breaking change over this?

Maybe we can make a new optional parameter instead of `onRetry` or to complement `onRetry`. Like `logRetry`?

jonasfj (changes requested):

retry/lib/retry.dart
drown0315: I think stackTrace can provide additional information about an exception, which can help developers pinpoint the specific cause of the exception. This is especially helpful when exceptions occur sporadically, or when the location of the exception differs each time you retry. The stack trace can assist developers in reproducing and fixing these difficult to diagnose exceptions.

retry/lib/retry.dart
drown0315: Updated.

retry/lib/retry.dart
drown0315: Yes, retryIf should not carry the stackTrace parameter.
Updated.

retry/lib/retry.dart
```diff
@@ -118,24 +118,38 @@ final class RetryOptions {
...
   /// as an [Exception].
+  ///
+  /// The [onRetry] `Function` below stands for one of two types:
+  /// - (Exception) -> FutureOr<void>
+  /// - (Exception, StackTrace) -> FutureOr<void>
   Future<T> retry<T>(
     FutureOr<T> Function() fn, {
     FutureOr<bool> Function(Exception)? retryIf,
-    FutureOr<void> Function(Exception)? onRetry,
+    Function? onRetry,
```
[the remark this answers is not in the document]
jonasfj: I wonder if there is an annotation or comment that'll allow us to hide the parameter from auto-completion :D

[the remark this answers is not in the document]
drown0315:
```
@Deprecated('Use `onRetryFailure` instead of `onRetry`')
FutureOr<void> Function(Exception)? onRetry,
FutureOr<void> Function(Exception, StackTrace)? onRetryFailure,
```
Thank for review. 

Do you think 'onRetryFailure' would be a good name for this?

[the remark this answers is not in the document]
jonasfj: Hmm, isn't logging the only legitimate use-case for `onRetry`, hence, `logRetry` might be a good name, or `retryLog` or???

We could also all it `whenRetry`.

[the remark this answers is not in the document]
drown0315: There are some names here:

* (i) `onFailure`.  regardless of whether Exception satisfies `retryIf`, it will be called.
* (ii) `onError`. any error on retry, it will be called.

Which one do you think is better?

PR #218, opened 2023-08-07
api_analysis_exploration: add r3 experiment

CicadaCinema:
This includes propagation of exports, so the summary produced should be fairly comprehensive.

One can analyse a local package to obtain a full summary, or a hosted package, in which case all appropriate versions are analysed and the differences displayed.

It remains to add a README description.

### DIFF
--- a/api_analysis/lib/r3/analyze.dart
+++ b/api_analysis/lib/r3/analyze.dart
@@ -379,7 +379,7 @@ extension on LibraryShape {
     for (final v in d.variables.variables) {
       final name = v.name.value();
       if (name is! String) {
-        throw fail('Unnamed variable $v');
+        fail('Unnamed variable $v');
       }
       define(VariableShape(
         name: name,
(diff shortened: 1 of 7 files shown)


CicadaCinema:
@jonasfj

api_analysis/lib/r3/analyze.dart
```diff
@@ -0,0 +1,491 @@
...
+        case TopLevelVariableDeclaration d:
+          shape.defineTopLevelVariable(d);
+      }
+    }
+  }
+
+  // TODO: is the opposite of 'transitive export', 'direct export'?
+  /// Iterate over the direct exports of this library and merge the set of
+  /// exported shapes of those libraries into this one.
+  /// Returns [true] if the shape of [library] has been changed as a result.
```
jonasfj, line 336: It might be cheaper to propagate around the `export` statements... so updating `library.exports` based on what the libraries being exported have in `exportedLibrary.exports`.

It's a bit more complicated, but this is why the `NamespaceFilter` has the merge/apply logic.

I think we can even be able to track which `LibraryShape` objects have had their `.exports` property modified in the first iteration. Because then we only need to further propogate those in following iterations (or something like that).


----------------

The idea with propogating the `export` declaration is that there are much fewer of them than there are of the `definedShapes`.

I think that if we propogate `export` declarations, we can create `exportedShapes` in a single pass afterwards.

note. maybe a good package to test this on would be something like `package:googleapis`, it's massive and adds a lot of symbols without breaking the API.

PR #223, opened 2023-09-28
sanitize_html: append rel attributes instead of overriding

isoos:


jonasfj:
Document why this is safe, it's not obvious to me. Is it? Do we currently pass-through rel? Is `rel` a safe thing to inject?

isoos:
>  Do we currently pass-through rel?

`rel` is currently an always-allowed attribute, so by default a pass-through: https://github.com/google/dart-neats/blob/master/sanitize_html/lib/src/sane_html_validator.dart#L133

However, if you have a return value from the `addLinkRel`, it overrides it. I think the current behavior should have either a `setLinkRel` callback, or the `addLinkRel` should append the attributes from the callback. In other words, I consider this a bug-fix, and not a new feature.

jonasfj (approved):

Issue #210, opened 2023-06-04
Could not start() after stop()


igokom:
* Actually got nothing to say more
 * Or is it not supposed to be used like that?

```dart
import 'package:neat_periodic_task/neat_periodic_task.dart';
import 'package:test/test.dart';

void main() {
  test('test task restart', () async {
    final scheduler = NeatPeriodicTaskScheduler(
      interval: Duration(seconds: 5),
      name: 'hello-world',
      timeout: Duration(seconds: 5),
      task: () async => print('*****Hello World********'),
      minCycle: Duration(seconds: 1),
    );

    scheduler.start();
    print("Start task");
    await Future.delayed(Duration(seconds: 6));
    await scheduler.stop();
    print("Stop task");
    scheduler.start();
    print("Start task again");
  });
}
```

jonasfj:
It is not intended to be restartable.

See: https://github.com/google/dart-neats/pull/226

--------------

Generally, I fear that ability to restart could lead to race conditions, making it unclear what state the scheduler is supposed to be in.

PR #226, opened 2023-10-17
Updated documentation comments

jonasfj:
Documentation update to fix https://github.com/google/dart-neats/issues/210

Context: `NeatPeriodicTaskScheduler` is intended to be used for background task that run for the duration of the process. This is generally aimed at the server space where the task state is persistent in a database or similar thing.

### DIFF
--- a/neat_periodic_task/lib/neat_periodic_task.dart
+++ b/neat_periodic_task/lib/neat_periodic_task.dart
@@ -115,7 +115,7 @@ class NeatPeriodicTaskScheduler {
   final Duration _minCycle;
   final Duration _maxCycle;
 
-  bool _running = false;
+  bool _started = false;
   final _stopping = Completer<void>();
   final _stopped = Completer<void>();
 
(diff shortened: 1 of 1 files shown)


neat_periodic_task/lib/neat_periodic_task.dart
```diff
@@ -186,6 +190,8 @@ class NeatPeriodicTaskScheduler {
   /// This returns a [Future] that is completed when the scheduler has finished
   /// any on-going iterations. This will not abort an ongoing task, but it will
   /// stop further iterations of the task.
+  ///
+  /// Once stopped a [NeatPeriodicTaskScheduler] cannot be started again.
   Future<void> stop() async {
```
sigurdm, line 195: Should we rename `stop()`->`close()` - seems more conventional, and would perhaps better indicate that you cannot restart.

sigurdm (approved):

PR #229, opened 2023-11-10
Allow relative URLs in the redirect location header.

### LINKED ISSUE #228: The updated RFC allows relative URLs in the `location` header, but the package rejects them.
I believe that the package used that method, because the original standard required absolute URLs in the location header:
https://www.rfc-editor.org/rfc/rfc2616#section-14.30

However, there seems to be an updated RFC, which now allows relative URLs in the location:
https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2

Source issue: https://github.com/dart-lang/pana/issues/1281

isoos:
Fixes #228.

### DIFF
--- a/safe_url_check/lib/safe_url_check.dart
+++ b/safe_url_check/lib/safe_url_check.dart
@@ -159,8 +159,10 @@ Future<bool> _safeUrlCheck(
   if (response.isRedirect &&
       response.headers[HttpHeaders.locationHeader]!.isNotEmpty &&
       maxRedirects > 0) {
+    final loc = Uri.parse(response.headers[HttpHeaders.locationHeader]![0]);
+    final nextUri = loc.hasScheme ? loc : url.resolveUri(loc);
     return _safeUrlCheck(
-      Uri.parse(response.headers[HttpHeaders.locationHeader]![0]),
+      nextUri,
       maxRedirects - 1,
       client,
       userAgent,
--- a/safe_url_check/lib/safe_url_check.dart
+++ b/safe_url_check/lib/safe_url_check.dart
@@ -160,7 +160,7 @@ Future<bool> _safeUrlCheck(
       response.headers[HttpHeaders.locationHeader]!.isNotEmpty &&
       maxRedirects > 0) {
     final loc = Uri.parse(response.headers[HttpHeaders.locationHeader]![0]);
-    final nextUri = loc.hasScheme ? loc : url.resolveUri(loc);
+    final nextUri = url.resolveUri(loc);
     return _safeUrlCheck(
       nextUri,
       maxRedirects - 1,
--- a/safe_url_check/lib/safe_url_check.dart
+++ b/safe_url_check/lib/safe_url_check.dart
@@ -39,12 +39,7 @@ import 'dart:async';
 
 import 'package:retry/retry.dart';
 
-import 'src/private_ip.dart';
-import 'src/unique_local_ip.dart';
-import 'src/version.dart';
-
-const _defaultUserAgent = 'package:safe_url_check/$packageVersion '
-    '(+https://github.com/google/dart-neats/tree/master/safe_url_check)';
+import 'src/safe_url_check.dart';
 
 /// Check if [url] is available, without allowing access to private networks.
 ///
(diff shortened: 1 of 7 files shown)


safe_url_check/lib/safe_url_check.dart
```diff
@@ -159,8 +159,10 @@ Future<bool> _safeUrlCheck(
   if (response.isRedirect &&
       response.headers[HttpHeaders.locationHeader]!.isNotEmpty &&
       maxRedirects > 0) {
+    final loc = Uri.parse(response.headers[HttpHeaders.locationHeader]![0]);
+    final nextUri = loc.hasScheme ? loc : url.resolveUri(loc);
```
sigurdm, line 163: Not sure, but maybe 
```suggestion
    final nextUri = loc.isAbsolute ? loc : url.resolveUri(loc);
```

sigurdm: From: https://api.dart.dev/stable/3.1.5/dart-core/Uri/isAbsolute.html

> A URI is an absolute URI in the sense of RFC 3986 if it has a scheme and no fragment.

safe_url_check/lib/safe_url_check.dart
```diff
@@ -159,8 +159,10 @@ Future<bool> _safeUrlCheck(
   if (response.isRedirect &&
       response.headers[HttpHeaders.locationHeader]!.isNotEmpty &&
       maxRedirects > 0) {
+    final loc = Uri.parse(response.headers[HttpHeaders.locationHeader]![0]);
+    final nextUri = loc.isAbsolute ? loc : url.resolveUri(loc);
```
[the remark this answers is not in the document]
isoos: I've prepared a server + redirect logic for a test, but the test logic checks for local / private IPs and it always fails for such tests. Any preference how to suppress the IP checks for tests?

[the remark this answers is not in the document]
jonasfj: Honestly, isn't it tempting to be lazy and just use: https://httpbin.org/redirect-to?url=status%2F200

Ignoring such local / private IP tests is kind of a bad idea, since the whole point is to avoid hitting private networks :D

I suppose one could do a one-off using a constant like `bool.fromEnvironment()` but that doesn't make testing easier, since you'll have to do `-D ..` something.

Let's either not test it, or use httpbin to not make our problems bigger than they have to be.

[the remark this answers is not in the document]
isoos: > Honestly, isn't it tempting to be lazy and just use: https://httpbin.org/redirect-to?url=status%2F200

Already added that: https://github.com/google/dart-neats/pull/229/files#diff-d7c4d80be7934367cf516c2c7d1c65dd444fc2a10786ea20790381d3c361dea1R30

> Ignoring such local / private IP tests is kind of a bad idea, since the whole point is to avoid hitting private networks :D

I think it is not unreasonable to turn off some parts of the library for local testing - it feels that more coverage is better in this case. I'll check if I can implement something that is not exposed to the end user in any other way.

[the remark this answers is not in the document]
isoos: Updated the PR with refactor + override for local testing. I can remove the commit if needed, but I think it is better to have it this way.

safe_url_check/lib/src/safe_url_check.dart
```diff
@@ -0,0 +1,144 @@
...
+  final RetryOptions retryOptions;
+  final Duration timeout;
+  final bool skipLocalNetworkCheck;
+
+  SafeUrlChecker({
+    required this.client,
+    required this.userAgent,
+    required this.retryOptions,
+    required this.timeout,
+    this.skipLocalNetworkCheck = false,
```
jonasfj, line 80:
```suggestion
    @visibleForTestingOnly
    this.skipLocalNetworkCheck = false,
```

safe_url_check/lib/src/safe_url_check.dart
```diff
@@ -0,0 +1,144 @@
...
+      if (client == null) {
+        c.close(force: true);
+      }
+    }
+  } on Exception {
+    return false;
+  }
+}
+
+class SafeUrlChecker {
```
jonasfj, line 68: Why are we creating a class that has no purpose?

This is just a convoluted way to make function, right?

Why not just keep the old `_safeUrlCheck` make it `safeUrlCheck` in the `safe_url_check/lib/src/safe_url_check.dart` file and add a parameter `skipLocalNetworkCheck` which defaults to `false` and has `@visibleForTestingOnly` annotation.

Or we could just accept that we don't have testing, or use httpbin.org for testing.

jonasfj: Or put the parameter on `doSafeUrlCheck`.

This introduces a class `SafeUrlChecker` which we only every construct inorder to call a single method.

It's just a confusing way to have a function that you can pass arguments to in two different ways.

Another, somewhat simpler and ugly option is to make a global:
```dart
@visibleForTestingOnly
bool skipLocalNetworkCheck = false;
```

Put this in an `lib/src/...` file and check it from within `safeUrlCheck`.
If no code actively sets this property, hopefully the compiler will optimize it away.

safe_url_check/lib/src/safe_url_check.dart
isoos: This will pull in `package:meta` as a dependency. I don't mind, just wanted to note.

safe_url_check/lib/src/safe_url_check.dart
isoos: > This is just a convoluted way to make function, right?

I didn't like the recursive function with 6+ parameters. The class keeps the non-changing parts as fields.

isoos: Updated, PTAL.

jonasfj (approved):

PR #231, opened 2024-01-04
fetch_package initial commit

sigurdm:
A proposal for a small utility...

### DIFF
--- a/fetch_package/README.md
+++ b/fetch_package/README.md
@@ -1,0 +1,33 @@
+# fetch_package
+
+A utility for easily downloading and extracting a package archive from pub.dev
+or other repository.
+
+Useful for reviewing the source-code of a pub package or for easily running
+examples.
+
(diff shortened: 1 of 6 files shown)


jonasfj (approved):
LGTM, if we want to ship it this way.

fetch_package/README.md
```diff
@@ -0,0 +1,33 @@
...
+This package is intended for global installation. Run this to install the latest version:
+
+```
+> dart pub global activate fetch_package # Only on first run
+```
+
+If you want to inspect the source code of the latest version of this package:
+
+```
+> dart pub global run fetch_package fetch_package
```
jonasfj, line 20: I'd suggest using a different package as example...

```suggestion
> dart pub global run fetch_package retry
```

And maybe show something like:
```
> dart pub global run fetch_package <package> [version]
```

PR #238, opened 2024-05-23
Use pub workspaces

sigurdm:
And upgrade dependencies.

This is mostly a demonstration. It might be wise to wait a bit before landing.


jonasfj:
Migrated outside this PR

PR #239, opened 2024-05-31
update dartdoc_test

takumma:


dartdoc_test/example/example.dart
```diff
@@ -0,0 +1,109 @@
...
+  /// Throws:
+  /// - [ArgumentError] if [n] is negative.
+  int factorial(int n) {
+    if (n < 0) throw ArgumentError('Negative numbers are not allowed.');
+    return n == 0 ? 1 : n * factorial(n - 1);
+  }
+}
+
+/// Utility class for various static methods.
+class Utility {
```
jonasfj, line 58: I know this is an example, but this kind of class is rather uncommon in dart -- or rather it ought to be :see_no_evil: 

static functions should just be top-level, if there is no reason for them to be static.

It might be more interesting to just make `gcd` and `isPalindrome` top-level functions.

If we want something fun with classes we should try to make subclass instead :D

dartdoc_test/lib/src/resource.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:analyzer/file_system/file_system.dart';
+import 'package:analyzer/file_system/memory_file_system.dart';
+import 'package:analyzer/file_system/physical_file_system.dart';
+import 'package:path/path.dart';
+
+class CompositeResourceProvider implements ResourceProvider {
```
jonasfj, line 6: I'd consider just using the OverlayResourceProvider on top of PhysicalResourceProvider.

Because if you get a `Folder` object from the memoryProvider, then calling [getChild](https://pub.dev/documentation/analyzer/5.13.0/file_system_file_system/Folder/getChild.html) might not return the correct result.

Correctly wrapping a `ResourceProvider` is not easy, if you want to try you can dive into the implementation for `OverlayResourceProvider`. But just using `OverlayResourceProvider` is probably fine. We don't need to inject many files anyways :D

dartdoc_test/example/example.dart
takumma: thank you for reviewing. I fix it.

dartdoc_test/lib/src/resource.dart
takumma: I replace it to just use [`OverlayResourcceProvider`](https://pub.dev/documentation/analyzer/5.13.0/file_system_overlay_file_system/OverlayResourceProvider-class.html). I believe a single `resourceProvider` instance should be shared within `dartdoc_test`, but I'm not sure if `analysisContext` or `session` should be share as well... I think I need to understand the `package:analyzer` and anlysis process.

PR #242, opened 2024-06-17
[dartdoc_test] add documentation samples analyzer

takumma:
- including suggestions from review of  #241
  - add License headers to all files.
  - add LF/CRLF line breaks tests.
  - add documentation comments.
- implement analysis.


dartdoc_test/test/extractor_test.dart
```diff
@@ -47,5 +61,31 @@ final x = 1;
...
+      });
+
+      test('can strip LF (linux) line breaks', () {
+        final actualText =
+            '/// This is a linux line break comment.\n/// ```dart\n/// final x = 1;\n/// ```';
+        final expectText = '''
+This is a linux line break comment.
+```dart
+final x = 1;
+```''';
```
jonasfj, line 85: nit: Consider writing it on the same form as the `actualText` that way it becomes super easy to compare what charaters changed.
```suggestion
        final expectText = '''This is a linux line break comment.\n```dart\nfinal x = 1;\n```''';
```

dartdoc_test/lib/src/resource.dart
```diff
@@ -1,15 +1,61 @@
...
+// Directory to store generated code samples
+final testDirectory = Directory('${_current.path}/.dartdoc_test');
+
+final _contextCollection = AnalysisContextCollection(
   resourceProvider: resourceProvider,
   includedPaths: [_current.absolute.path],
-).contextFor(_current.absolute.path);
+);
+
+final currentContext = _contextCollection.contextFor(_current.absolute.path);
```
jonasfj, line 37: I'd avoid making stateful global variables like `currentContext`, `testDirectory`, `resourceProvider`.

Consider making a `DartDocTestContext` and placing all the state you need on that. Or something like that. Could also be called a `TestContext` just to keep name short and simple :D

jonasfj (approved):

dartdoc_test/test/extractor_test.dart
takumma: I fixed to so. thank you :)

dartdoc_test/lib/src/resource.dart
takumma: I created a singleton class called `TestContext`.

dartdoc_test/example/README.md
```diff
@@ -0,0 +1,2 @@
+A sample command-line application with an entrypoint in `bin/`, library code
+in `lib/`, and example unit test in `test/`.
```
chinmoy12c, line 2: I'm not sure I understand this `README`. Can we have a proper README explaining, how to setup and test things?

chinmoy12c (approved):
Can we ad

dartdoc_test/example/README.md
takumma: It is a example app's README, but I made it better :)

dartdoc_test/example/test/dartdoc_test.dart
```diff
@@ -14,7 +14,6 @@
 
 import 'package:dartdoc_test/dartdoc_test.dart';
 
-Future<void> main(List<String> args) async {
-  final dartdocTest = DartDocTest(DartDocTestOptions.parse(args));
-  dartdocTest.analyze();
+void main() {
```
jonasfj, line 17:
```suggestion
void main() => runDartdocTest();
```

PR #243, opened 2024-06-24
Allow 'OPENSSH PRIVATE KEY'

RickPoleshuck:
Allow parsing .ssh/id_rsa files that use 'OPENSSH PRIVATE KEY'.

### DIFF
--- a/pem/lib/pem.dart
+++ b/pem/lib/pem.dart
@@ -111,6 +111,12 @@ enum PemLabel {
   /// See [section 10 RFC 7468](https://tools.ietf.org/html/rfc7468#section-10).
   privateKey,
 
+  /// Unencrypted PKCS #8 Private Key Information Syntax structure
+  /// (`'OPENSSH PRIVATE KEY'`).
+  ///
+  /// See [section 10 RFC 7468](https://tools.ietf.org/html/rfc7468#section-10).
+  openSshPrivateKey,
+
   /// Encrypted PKCS #8 Private Key Information Syntax structure
   /// (`'ENCRYPTED PRIVATE KEY'`).
   ///


pem/lib/pem.dart
```diff
@@ -111,6 +111,12 @@ enum PemLabel {
   /// See [section 10 RFC 7468](https://tools.ietf.org/html/rfc7468#section-10).
   privateKey,
 
+  /// Unencrypted PKCS #8 Private Key Information Syntax structure
+  /// (`'OPENSSH PRIVATE KEY'`).
+  ///
+  /// See [section 10 RFC 7468](https://tools.ietf.org/html/rfc7468#section-10).
```
jonasfj, line 117: I don't see anything mentioning `"OPENSSH PRIVATE KEY"` in:
https://datatracker.ietf.org/doc/html/rfc7468#section-10

`unsafeIgnoreLabel` exists for things outside the spec.

jonasfj (changes requested):

jonasfj:
This package is about parsing PEM keys (not PEM-like things).

Mainly, it makes it easy to parse out the base64 bytes correctly, and discourages parsing the wrong kind of key, by forcing you to specify a label.

Anything else is outside the scope of this package, "OPENSSL PRIVATE KEYS" are probably not PEM keys. That's pretty obvious since the label isn't mentioned in the RFC.

See also: https://superuser.com/questions/1720991/differences-between-begin-rsa-private-key-and-begin-openssh-private-key

In general, I don't think we should add arbitrary labels that are outside the specification.

I do recognize that there might be a use case for parsing PEM-like structures with other labels. This is why `unsafeIgnoreLabel` exists.

If you want to contribute an example in the `README` for how to parse PEM-like structures that don't use the PEM labels. I think that'd be useful to have another example section at the bottom of the README.

------------------

In practice I think it can also be argued that one should export the key in the correct format:
https://man.openbsd.org/ssh-keygen#m

PR #246, opened 2024-07-14
`retry`: support async stream.

nrbnlulu:


retry/lib/retry.dart
```diff
@@ -143,6 +143,38 @@ final class RetryOptions {
       await Future.delayed(delay(attempt));
     }
   }
+
+
+
+Stream<T> stream <T>(
```
isoos, line 149: If we keep this, it may only work for a subset of stream categories. Some streams need to be restarted from scratch when an error happens, but then in such cases the result stream is not aware that the restart happened and it will get repeated items.

In practice we've found that retrying the entire operation that encloses the stream processing is more robust.

nrbnlulu: Can you wrap a stream in a way that you can consume it on the fly with simple retry?

isoos: The typical request-response protocol (like HTTP) is best abstracted by `retry` on each request. Retrying the underlying stream doesn't make sense here, esp. if the operation is mutating a state.

nrbnlulu: hmm I thought it would be good for WebSockets but fair enough...
Thanks for your time.

PR #247, opened 2024-07-21
[dartdoc_test] implement Reporter

takumma:
previous PR: #245 

- Update logging
  - create Reporter (for test and stdout). And update to use Reporter.
    - Please refer integration test output file (`dartdoc_test.txt` and `run analyze command.txt`) to see what the output of ReporterForTest and ReporterForStdout` will look like.
  - implement ANSI, Summary log.
  - print relative path
- update test
  - add output directory option.
  - create a new extractor integration test. it generates sample code dart files to `testdata` directory
  - add integration test for `runDartdocTest`.


dartdoc_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'package:source_span/source_span.dart';
+import 'package:test/test.dart';
+
+abstract class Reporter {
+  void reportSourceFile(String filename, void Function(SourceFileReporter) fn);
+}
+
+abstract class SourceFileReporter {
+  void reportCodeSample(
+      String identifier, int number, void Function(CodeSampleReporter) fn);
```
jonasfj, line 24: I'd suggest documenting this a bit more... like what is `identifier` ? and `number`?

dartdoc_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,76 @@
...
+  void reportSourceFile(String filename, void Function(SourceFileReporter) fn) {
+    group(filename, () {
+      fn(this);
+    });
+  }
+
+  @override
+  void reportCodeSample(
+      String identifier, int number, void Function(CodeSampleReporter) fn) {
+    test('$identifier-($number)', () {
```
jonasfj, line 61:
```suggestion
    test('$identifier ($number)', () {
```

dartdoc_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,76 @@
...
+      fn(this);
```
jonasfj, line 62: you need to clear `issues` first, or make a new `CodeSampleReporter` object each time...

dartdoc_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,76 @@
...
+abstract class SourceFileReporter {
+  void reportCodeSample(
+      String identifier, int number, void Function(CodeSampleReporter) fn);
+}
+
+abstract class CodeSampleReporter {
+  void issues(SourceSpan span, String message);
+}
+
+final class ReporterForStdout
```
jonasfj, line 31: not important, this just my personal preference for structure.

I would do:
```
final class _ReporterForStdout implements ... { ... }
final class _RepoterForTest implements ... { ... }
```
So that they are private, then I would make static methods on `Reporter`, like:
```
abstract class Reporter {
   ...

   /// Create a reporter that prints to stdout
   static Reporter stdout() => _ReporterForStdout();

   /// Create a reporter that registers tests in `package:test`
   static Reporter test() => _RepoterForTest();
}
```

Again, this is just my personal preference. I prefer this because when someone goes to read the generated documentation (created by dartdoc), they won't see all these classes that they don't need to know about. They don't need to know about `RepoterForTest`.

If someone was to write a function that takes a `Reporter` as argument, they should accept `Reporter`, if they accidentally the `ReporterForStdout` instead, then that may work if they only need to output for stdout. But it'll stop working for other scenarios.

If someone writes a function that takes a reporter as argument, example: `void myFunction(Reporter r) {...` that's great. But I don't want them to be able to do `if (r is RepoterForTest) {...` and special-case their test dependent on the kind of `Reporter` they are given.

I'm guessing that nobody will actually write such functions, but we never know. Mostly, when making APIs I strive to make sure that users only have access to the things they need, and that everything they can see (or access) as a meaning.

Again, this is just like my opinion, it's hard to say that there is a right and a wrong here. There are many ways to do things.

dartdoc_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:source_span/source_span.dart';
+import 'package:test/test.dart';
+
+abstract class Reporter {
```
jonasfj, line 18: Perhaps we should consider:
```suggestion
abstract base class Reporter {
```
That way, we force everybody who wants to make a custom `Reporter` to do: `class MyCustomReporter extends Reporter {....`.

At this point in time this doesn't really change anything. But if in the future we wanted to add another method to `Reporter`, it would be a breaking change! We'd have to bump the major version on `package:dartdoc_test`. Unless we force everybody to use `extends Reporter` instead of `implements Reporter`.

**Example**: If in the future we decide we want to add `reportInvalidSourceFile(String filename, String errorMessage)` method to `Reporter`. If we did:
```dart
abstract base class Reporter {
  void reportSourceFile(String filename, void Function(SourceFileReporter) fn);
  void reportInvalidSourceFile(String filename, String errorMessage);
}
```

Then obviously, it'd be a breaking change, because everybody who made a custom reporter like: `class MyCustomReporter extends Reporter {....` wouldn't have an implementation of `reportInvalidSourceFile`.

But if instead we did:
```dart
abstract base class Reporter {
  void reportSourceFile(String filename, void Function(SourceFileReporter) fn);
  void reportInvalidSourceFile(String filename, String errorMessage) {
    // ignore
  }
}
```

Then it wouldn't be a breaking change. Custom reporters like `class MyCustomReporter extends Reporter {....` will continue to work, the new kind of reports about invalid source files will be ignored by `MyCustomReporter`, but since this is a new kind of message, it's might not be important to it.

Thus, we'd be able to roll out a new kind of field or method, so long as it has a default implementation. Though this only works if people who make custom reporters use `extends Reporter`, and not `implements Reporter`. However, we can always force people to use `extends Reporter` by making it an `abstract base class Reporter`

dartdoc_test/lib/src/logger.dart
```diff
@@ -12,15 +12,75 @@
...
+final class Logger {
+  final LogFunction _log;
+  Logger(this._log);
+
+  void info(String message) => _log(message, LogLevel.info);
+  void debug(String message) => _log(message, LogLevel.debug);
+  void warning(String message) => _log(message, LogLevel.warning);
+  void error(String message) => _log(message, LogLevel.error);
+  void log(String message, LogLevel level) => _log(message, level);
+}
```
jonasfj, line 52: Just a crazy idea, feel free to ignore, but we could also do:
```suggestion
extension Logger on LogFunction {
  void info(String message) => this(message, LogLevel.info);
  void debug(String message) => this(message, LogLevel.debug);
  void warning(String message) => this(message, LogLevel.warning);
  void error(String message) => this(message, LogLevel.error);
  void log(String message, LogLevel level) => this(message, level);
}
```

dartdoc_test/lib/src/command/analyze.dart
```diff
@@ -0,0 +1,87 @@
...
+    ));
+    logger.info('Extracting code samples ...');
+    await dartdocTest.extract();
+
+    logger.info('Analyzing code samples ...');
+    final result = await dartdocTest.analyze();
+
+    for (final r in result) {
+      if (r.errors.isEmpty) {
+        logger.info('No issues found!');
```
chinmoy12c, line 58: Won't this print `No issues found!` for every result?

dartdoc_test/lib/src/extractor.dart
```diff
@@ -50,14 +52,19 @@ class DocumentationCodeSample {
     this.imports = const [],
   });
 
-  String get wrappedCode {
-    final fileName = comment.span.sourceUrl;
+  String wrappedCode(Directory testDir) {
+    var fileName = comment.span.sourceUrl;
```
chinmoy12c, line 56: We can probably use `final` here.

dartdoc_test/lib/src/command/analyze.dart
```diff
@@ -0,0 +1,87 @@
...
+      }
```
jonasfj, line 59:
```suggestion
    final hasIssues = result.any((r) => r.errors.isNotEmpty);
    if (!hasIssues) {
      logger.info('No issues found!');
    }
    for (final r in result) {
```

dartdoc_test/lib/src/extractor.dart
takumma: Thank you for reviewing. I applied your suggetions.

dartdoc_test/test/testdata/run analyze command.txt
```diff
@@ -1,17 +1,22 @@
...
-line 93, column 11 of example/lib/example.dart: Local variable 'gcd' can't be referenced before it is declared.
+example/lib/example.dart:79:36: : Local variable 'gcd' can't be referenced before it is declared.
    ╷
-93 │ /// final gcd = gcd(48, 18);
-   │           ^^
+79 │ /// Should return: Local variable 'gcd' can't be referenced before it is declared.
+   │                                    ^^^
    ╵
 
+FAILED: 28 issues found (Found 11 code samples in 3 files)
```
jonasfj, line 22: the output only shows 3, is that intentional?

dartdoc_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,186 @@
...
+        if (issue.commentSpan != null) {
+          fail(issue.commentSpan!.info(issue.message));
+        }
+      },
+      // skip: !_verbose && issue.generatedSpan == null,
+    );
+  }
+}
+
+class Issue {
```
jonasfj, line 148: Should this be final? or do you intent for people using the API to make their own subclasses?

jonasfj (approved):

PR #248, opened 2024-08-05
[dartdoc_test] add some ways to ignore analysis

takumma:
add 2 ways to ignore code sample analysis.
- Ignore with tag `#no-test`
  - I implemented to ignore a code sample if `dart#no-test` is specified as the language of code block.
- ignore file with exclude option
  - I added `exclude` option to cli and test function. so we can ignore files or directories.

and I add public documentation and do some work to get it published!

```dart
/// ```dart#no-test
/// func() // it will not be analyzed.
/// ```
```


dartdoc_test/example/lib/example.dart
```diff
@@ -52,53 +64,11 @@ int multiply(int x, int y) {
...
-  while (b != 0) {
-    var t = b;
-    b = a % b;
-    a = t;
-  }
-  return a;
-}
-
-/// Convert a string to a list of lines.
+/// // dartdoc_test:ignore_error
```
jonasfj, line 70: if we can make this work, because they it won't show up in the generated dartdoc documentation.
```suggestion
// dartdoc_test:ignore_error
```

jonasfj: Maybe, we could do this, but it not as pretty.
```
/// my summary
///
/// <!-- ignore:dartdoc_test -->
/// my example:
/// ```dart
/// ...
/// ```
```

dartdoc_test/lib/src/extractor.dart
```diff
@@ -45,17 +46,24 @@ class DocumentationComment {
...
   final String code;
-  final List<String> imports;
 
   DocumentationCodeSample({
     required this.comment,
     required this.code,
-    this.imports = const [],
   });
 
+  bool get ignore => code.contains('// dartdoc_test:ignore_error');
```
jonasfj, line 62:
```suggestion
  bool get ignore => code.contains('// ignore:dartdoc_test');
```

jonasfj: TODO: investigate if we can look for this outside the _documentation comment_, which might be very hard :D

dartdoc_test/README.md
```diff
@@ -1,32 +1,82 @@
...
 void main() => runDartdocTest();
+
 ```
 
-and then
+# Ignore analysis
 
-```bash
-dart test
+If you don't want to analyze a particular code sample, you can exclude it by adding an `// dartdoc_test:ignore_error` comment within the code sample.
```
jonasfj, line 56: outdated?

dartdoc_test/README.md
```diff
@@ -1,32 +1,82 @@
...
+
+````
+/// This code sample will not be analyzed.
+///
+/// ```dart
+/// // dartdoc_test:ignore_error
```
jonasfj, line 62: does this work?

dartdoc_test/lib/src/model.dart
```diff
@@ -0,0 +1,133 @@
...
+// limitations under the License.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:source_span/source_span.dart';
+
+/// A dart file that contains documentation comments and code samples.
+class DartdocTestFile {
```
jonasfj, line 22: Consider if maybe this class can be `final`?

dartdoc_test/lib/src/model.dart
```diff
@@ -0,0 +1,133 @@
...
+/// A code sample extracted from a documentation comment.
+class DocumentationCodeSample {
+  /// The documentation comment that contains the code sample.
+  final DocumentationComment comment;
+
+  /// content of the code sample.
+  final String code;
+
+  /// Whether the code sample has a `no-test` tag.
+  final bool noTest;
```
jonasfj, line 80:
```suggestion
  /// Whether the code sample has a `no-test` tag.
  ///
  /// ```text
  /// /// ```dart#no-test
  /// /// // this dart-code is not tested
  /// /// ```
  /// ```
  final bool noTest;
```

dartdoc_test/lib/src/model.dart
takumma: I made some classes to use `final`.

dartdoc_test/README.md
takumma: Thank you for suggestion! I updated it!

dartdoc_test/README.md
```diff
@@ -1,32 +1,79 @@
...
 
-You can test documentation code samples by creating `test/dartdoc_test.dart`.
+By creating `test/dartdoc_test.dart`, you can test code samples with using the `dart test` command.
+
+```bash
+# Create test file for code samples to "test/dartdoc_test.dart"
+dart run dartdoc_test add
+
+# Run test command
+dart test
```
jonasfj, line 35:
```suggestion
# Add dartdoc_test as dev dependency
dart pub add dev:dartdoc_test

# Create test file for code samples to "test/dartdoc_test.dart"
dart run dartdoc_test add

# Run test command
dart test
```

jonasfj (approved):
r+

dartdoc_test/test/testdata/code_sample/error_example_0.dart
jonasfj, line 1: I don't think we can make `dart format` ignore these files.
We can either disable `dart format` check in `mono_pkg.yaml` (which is a bit sad), or we can use a different extension here :rofl: 

Like `.dart-example` instead of `.dart` :see_no_evil: 

I'm not sure I have any good ideas :/

takumma: I found this and, as a temporary solution, I rewrote the example so that it would not be subject to format checks.

```
/// 1 positional argument expected by 'factorial', but 0 found.
/// ```dart
/// final fact = factorial();
/// print(fact); // 120
/// ```
```

PR #259, opened 2024-10-23
Additional allow attributes & tags

dab246:


jonasfj:
At the moment this package is aimed at aligning with github gfm sanitization rules:
https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb

Arguably, these have changed, I don't think github uses that code anymore.

But I'm hesitant to just add features. If there is a lot of community members that would rather have an HTML sanitation package with more advanced features I'm inclined to suggest that you write such a package, publish and maintain it yourself.

Feel free to fork this package and give it a new better name, ideally collaborate with others.
I see that @jawad111 has suggestions in https://github.com/google/dart-neats/pull/263.

Honestly, I'd be more than happy for `package:sanitize_html` to have a link in the `README.md` telling users that if they want advanced features they should consider one of the following packages... (assuming you make a good package obviously :rofl:)

----

On topic of this PR, I might get around to doing a review, but again, I'm hesitant to accept more features. Maybe, if it turns out we need them elsewhere too.

dab246:
> At the moment this package is aimed at aligning with github gfm sanitization rules: https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb
> 
> Arguably, these have changed, I don't think github uses that code anymore.
> 
> But I'm hesitant to just add features. If there is a lot of community members that would rather have an HTML sanitation package with more advanced features I'm inclined to suggest that you write such a package, publish and maintain it yourself.
> 
> Feel free to fork this package and give it a new better name, ideally collaborate with others. I see that @jawad111 has suggestions in #263.
> 
> Honestly, I'd be more than happy for `package:sanitize_html` to have a link in the `README.md` telling users that if they want advanced features they should consider one of the following packages... (assuming you make a good package obviously 🤣)
> 
> On topic of this PR, I might get around to doing a review, but again, I'm hesitant to accept more features. Maybe, if it turns out we need them elsewhere too.

Yes, Maybe we should develop a separate package that allows people to use the advanced features of sanitize_html. If I have time, I will do it. Stay tuned in this PR.

PR #270, opened 2025-03-26
Ignore StateError when closing sockets, and attempt multiple times.

jonasfj:
I think this will fix issues where the connections are not closed correctly.

### DIFF
--- a/neat_cache/lib/src/providers/resp.dart
+++ b/neat_cache/lib/src/providers/resp.dart
@@ -145,7 +145,7 @@ class RespClient {
   /// [1]: https://redis.io/topics/protocol
   Future<Object?> command(List<Object> args) async {
     if (_closing) {
-      throw RedisConnectionException._('redis connection is closed');
+      throw RedisConnectionException._('redis connection is closing');
     }
 
     final out = BytesBuilder(copy: false);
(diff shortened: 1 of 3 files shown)


isoos (approved):

neat_cache/lib/src/providers/resp.dart
```diff
@@ -215,13 +215,39 @@ class RespClient {
...
+          // ignore
+        }
       });
       await _closed.future;
     } else {
-      await _output.close().catchError((_) {/* ignore */});
+      try {
+        // NOTE: Sadly, this can throw, it can even throw synchroniously,
+        //       implying [Socket] from 'dart:io' fails to implement
+        //       [StreamSink] as specified.
```
sigurdm, line 246:
```suggestion
        // NOTE: Sadly, this can throw, it can even throw synchroniously,
        //       implying [Socket] from 'dart:io' fails to implement
        //       [StreamSink] as specified.
```
If this is true we should add a issue number here.

neat_cache/lib/src/providers/resp.dart
```diff
@@ -215,13 +215,39 @@ class RespClient {
...
+          await _output.close();
+        } catch (e) {
+          // ignore
+        }
       });
       await _closed.future;
     } else {
-      await _output.close().catchError((_) {/* ignore */});
+      try {
+        // NOTE: Sadly, this can throw, it can even throw synchroniously,
```
sigurdm:
```suggestion
        // TODO(https://github.com/dart-lang/sdk/issues/61743): Sadly, this can throw, it can even throw synchroniously,
```

PR #273, opened 2025-04-29
Move from mono_repo to pub workspaces

jonasfj:


dartdoc_test/lib/src/extractor.dart
```diff
@@ -45,17 +45,15 @@ List<DocumentationComment> extractDocumentationComments(ParsedUnitResult r) {
...
-      final content = stripComments(span.text);
-
-      comments.add(DocumentationComment(
-        path: r.path,
-        contents: content,
-        span: span,
-        imports: imports,
-      ));
-    }
+    final span = file.span(comment.offset, comment.end);
```
sigurdm, line 48: Is this part of the migration?

sigurdm (approved):

PR #276, opened 2025-05-05
Fixed Builder in package:typed_sql

jonasfj:


typed_sql/lib/builder.dart
```diff
@@ -55,23 +53,7 @@ final class _TypedSqlBuilder extends g.Generator {
     g.LibraryReader targetLibrary,
     BuildStep buildStep,
   ) async {
-    // Look up CustomDataType for later use!
-    final typedSqlLibrary = await targetLibrary.element.session
```
jonasfj, line 59: This was a bit of a hack, because I can't really get the `AnalysisSession` from other places.

Somehow, I'm actually able to get the library element, but the `exportNamespace` is empty! Even in the resolved library.

I don't know, I think `package:build` is playing games with me or something, maybe they don't want to be able to see things this way.

The new approach is a bit weird, but I don't have to do the lookup, up-front, and hopefully it works too.

sigurdm (approved):

sigurdm:
Should this have a test?

jonasfj:
We do have tests, but build logic behaved differently when referenced in a different project.

PR #287, opened 2025-05-20
Support for postgres BC datetimes, notably `DateTime.utc(0)`

jonasfj:
It's not that 'BC' dates are frequently used, or that `DateTime.utc(0)` is important, it's that `DateTime.utc(0)` or `DateTime(0)` can be nice to use instead of `NULL`.


typed_sql/lib/src/adapter/postgres_adapter.dart
```diff
@@ -451,6 +451,9 @@ final class _PostgresRowReader extends RowReader {
 
 List<Object?> _paramsForPostgres(List<Object?> params) => params
     .map((p) => switch (p) {
+          // Negative years and year zero most be written as BC in postgres
```
sigurdm, line 454: Perhaps add a reference to some docs or discussion?

sigurdm (approved):

PR #290, opened 2025-05-22
Prepare typed_sql version 0.1.3

jonasfj:


typed_sql/lib/src/adapter/adapter.dart
```diff
@@ -12,11 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// @docImport 'package:test/test.dart';
-library;
-
 import 'dart:typed_data';
-
+;
```
sigurdm, line 16:
```suggestion
```

typed_sql/lib/src/adapter/adapter.dart
```diff
@@ -195,7 +192,7 @@ abstract base class DatabaseAdapter extends Executor {
   /// > compatibility guarantees.
   ///
   /// This is intended for debugging purposes, and pairs nicely with
-  /// [printOnFailure] from `package:test/test.dart`.
+  /// `printOnFailure` from `package:test/test.dart`.
```
sigurdm, line 195: Did the docimport not work?

pubspec.lock
```diff
@@ -13,10 +13,10 @@ packages:
     dependency: transitive
     description:
       name: analyzer
-      sha256: "13c1e6c6fd460522ea840abec3f677cc226f5fec7872c04ad7b425517ccf54f7"
+      sha256: "904ae5bb474d32c38fb9482e2d925d5454cda04ddd0e55d2e6826bc72f6ba8c0"
```
sigurdm, line 16: Are you upgrading here on purpose?

sigurdm (approved):

typed_sql/lib/src/adapter/adapter.dart
jonasfj: On pub.dev we don't resolve `dev_dependencies`, so analysis will fail, but also rendered dartdoc won't have the links.

sigurdm: Ah yeah - didn't think about that...

PR #298, opened 2025-10-29
Better message when a 'Row' class is not included in the `build.yaml` sources.

isoos:
This affects schema definitions that span over multiple files, but their `build.yaml` sources list only includes only the main library. In such cases the `T must be a class in `Table<T>` properties` message was not really helpful.

### DIFF
--- a/typed_sql/CHANGELOG.md
+++ b/typed_sql/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.1.5
  * Upgraded `package:build` dependency to allow `>=3.0.0 <5.0.0`.
+ * Added check for a `Row` class not included in the `build.yaml` sources.
 
 ## 0.1.4
  * Upgraded dependencies (incl. `package:analyzer`).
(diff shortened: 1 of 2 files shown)


isoos:
> Does this deserve a test?

I think this is more like a nuance - though it did take me some time to figure out what's going on with the old error message. The package doesn't seem to have similar end2end tests, we should probably start adding, maybe outside of this PR?

jonasfj:
Okay, that's confusing.

Perhaps it's unwise to use a custom set of sources in build.yaml.

I thought that the analyzer in a build step would have access to all files,
not just the ones listed on build.yaml

We should read and understand this in pkg:build documentation.

I'm guessing a similar resolution issue could occur many places when
parsing schema, tables, etc.
Granted row is probably the most likely candidate.

--
Regards Jonas Finnemann Jensen.

tors. 30. okt. 2025, 14.53 skrev István Soós ***@***.***>:

isoos:
> I thought that the analyzer in a build step would have access to all files, not just the ones listed on build.yaml

This line first only had schema.dart in it, and it didn't recognized the `Task` class.
https://github.com/dart-lang/pub-dev/pull/9045/files#diff-d65142b04777a89292faa807caca594aa6765d8d3b0202345b4a80353f6cb273R11

Issue #306, opened 2025-11-26
[retry] Add ability to configure next timeout delay factor


vasilich6107:
* Currently `retry` doubles the timeout, 
 * it would be even more flexible if we were able to configure to arbitrary value

https://github.com/google/dart-neats/blob/master/retry/lib/retry.dart#L108

PR #307, opened 2025-11-26
[retry] configurable exponent

vasilich6107:
Fixes https://github.com/google/dart-neats/issues/306

### DIFF
--- a/retry/CHANGELOG.md
+++ b/retry/CHANGELOG.md
@@ -1,3 +1,6 @@
+## v3.1.3
+ * Expose ability to configure `exponent`
+
 ## v3.1.2
  * Make the `RetryOptions` class `final`.
 
--- a/retry/lib/retry.dart
+++ b/retry/lib/retry.dart
@@ -76,6 +76,9 @@ final class RetryOptions {
   /// Maximum number of attempts before giving up, defaults to 8.
   final int maxAttempts;
 
+  /// Exponent for the exponential backoff, defaults to 2.0.
+  final double exponent;
+
   /// Create a set of [RetryOptions].
   ///
   /// Defaults to 8 attempts, sleeping as following after 1st, 2nd, 3rd, ...,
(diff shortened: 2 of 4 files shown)


vasilich6107:
Hi @sigurdm 
Could help with PR review

sigurdm:
Maybe 'backoffMultiplicativeFactor': https://en.wikipedia.org/wiki/Exponential_backoff#:~:text=multiplicative%20factor%20or%20base%2C

jonasfj:
> The use case is straightforward - ability to retry with customizable backoff, it could be 1 to have just a number of retries without delay growing.

Is supplying `1` the primary use-case? I get that there may be use-cases where you want a fixed timeout, but why not using `maxDelay` then?

Curious, when in practice do you not want exponential back-off? For distributed systems (cloud) exponential back-off allows an overloaded resource to recover -- fixed retry interval is just bad design.

jonasfj:
> Could you clarify what I the main concern on adding a parameter to be configured?

There are outstanding PRs/issues to adding:
* `retryIfValue`
* `orElse`
* `onRetryFailure`
* `exponent`
* `onAttempt`

We already have a lot of options, do we need more? Or are we just entertaining use-case for 0.1% of users?

It's not like copying the code from this package into your library is going to cause code bloat or is difficult to do :D

But making a mistake in this package is hard to undo.

And adding new options incurs mental overhead for all users.

Also it's super easy to propose a new thing we should add, why there is a lot of PRs. The cost of adding a new thing is low, but that doesn't mean the mental overhead is low.

IMO, this package already suffers from the fact that `retryIf` is an extremely critical option to supply, but all the others are optional and not necessarily important for 99% of users.

(in hindsight I should probably have left most arguments as only exposed in `RetryOptions`.


> This is not some kind of a new functionality or breaking change.

That's true, and the only reason I feel we can reasonably entertain the proposal.

> While having the parameter in RetryOptions and not having it in function retry could be confusing to the user.
> Current API mirrors the payload of RetryOptions into retry fn.

Agreed, and I do wonder if the mirroring was a mistake.

-------

Part of all the outstanding PRs is also just that, I'm not sure if each feature would be used by 0.1% of users or 95% of users. And doing nothing might be safer than bloating the `retry` function with parameters users don't need.

Arguably, I also have limited bandwidth to spend on this package :D

Issue #309, opened 2025-11-29
[typed_sql] JSONB/JSON support


jonasfj:
* [x] https://github.com/google/dart-neats/pull/291
 * [x] Support for `@DefaultValue(JsonValue{'foo': 42, 'bar': null}))` on fields with `JsonValue`.
 * [x] Tests for `.distinct()` with `JsonValue`.
 * [x] Expressions and tests for `JsonValue`
   * [x] `.orElse` and `.orElseValue`
   * [x] `.isNull`, `.isNotNull`.
     * We need careful documentation that explains the different between `NULL` in SQL and `null` and JSON, at-least for operators that can work on `Expr<JsonValue?>`.
     * Including tests in `test/typed_sql/expr/`.
   * [x] JSON specific operators, like `.containsKey`, `.getKey(), `.length`, etc.
 * [x] Forbid use of `JsonValue` in:
    * [x] `@PrimaryKey`,
    * [x] `@ForeignKey` and `@References`
    * [x] `@Unique` constraints.
 * [x] Test use of `JsonValue` in `.groupBy`
 * [x] Test `JsonValue?` as field, see `test/typed_sql/nullable/`
 * [x] Add support and tests for `final class MyDataType implements CustomDataType<JsonValue>`
    * [ ] Forbid implementation of `CustomDataType<JsonValue>` and `Comparable<T>`.
 * [x] Update documentation guides to mention `JsonValue`.
 * [ ] Add dedicated documentation page with detailed example for `JsonValue`.
   * How to use `JsonValue`
     * Define a JSON field in a schema.
     * Declaring a default value.
     * Caveats when using distinct/group_by with `JsonValue` (possible difference between databases).
     * Using operators, and caveats around SQL `NULL` and JSON `null`.
       (Strongly recommend not using JSON `null`, if feasible).
     * Operators on `Expr<JsonValue>`, how to use, and caveats.
   * How to subclass using `CustomDataType<JsonValue>`
   * How to add custom operators on `Expr<CustomDataType<JsonValue>>` using JSON operators.

 @isoos

PR #311, opened 2025-12-03
Json distinct

jonasfj:
Fixing a bullet on https://github.com/google/dart-neats/issues/309


isoos (approved):
Note: maybe it would be worth to add a few tests for the `normalizeJson` to test it in isolation too.

PR #313, opened 2025-12-04
[retry] Add onAttempt callback to RetryOptions for logging

NJashwanth:
This PR for the retry package introduces a new optional onAttempt callback to the RetryOptions.retry() method. This callback is called on every retry attempt and receives the current attempt number. It is intended for logging, metrics, or instrumentation purposes without affecting the actual retry logic.

Existing behavior of fn() callbacks remains unchanged, ensuring full backward compatibility.

Example Usage
```
final r = RetryOptions(maxAttempts: 3);

await r.retry(
  () async {
    // Your retry logic here
  },
  onAttempt: (attempt) {
    print('Attempt #$attempt');
  },
);
```

### DIFF
--- a/retry/lib/retry.dart
+++ b/retry/lib/retry.dart
@@ -118,15 +118,27 @@ final class RetryOptions {
   /// If no [retryIf] function is given this will retry any for any [Exception]
   /// thrown. To retry on an [Error], the error must be caught and _rethrown_
   /// as an [Exception].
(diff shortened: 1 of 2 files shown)


jonasfj:
What is the use case?

Why isn't `onRetry` not good enough?

I'm personally, thinking that we easily have too many parameters on `retry` adding more will just make harder to use.

Also in general this isn't something you should care to do.

PR #362, opened 2026-04-30
Introduce `.insertValue()`

jonasfj:
Please **focus review on the first commit**, the second is updates to generated files.

This introduces `.insertValue` for convenience. 

I've hesitated to do this earlier because of the issue with `null` vs optional parameter being omitted.

I've narrowed it down such that:
 * Non-nullable fields that have a default value are optional named parameters (meaning parameter is nullable).
 * nullable fields that don't have a default value are optional named parameters (obviously the parameter is nullable).
 * nullable fields that have a default value are **required** named parameters (and users cannot insert the default value!).

### DIFF
--- a/typed_sql/lib/src/codegen/build_code.dart
+++ b/typed_sql/lib/src/codegen/build_code.dart
@@ -314,7 +314,7 @@ Iterable<Spec> buildTable(ParsedTable table, ParsedSchema schema) sync* {
         'Unsupported type "${field.typeName}"',
       ),
     };
-    if (field.backingType == field.typeName) {
+    if (!field.isCustomType) {
       return readBackingType;
     }
     return '''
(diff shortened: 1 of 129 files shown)


typed_sql/lib/src/codegen/build_code.dart
```diff
@@ -314,7 +314,7 @@ Iterable<Spec> buildTable(ParsedTable table, ParsedSchema schema) sync* {
         'Unsupported type "${field.typeName}"',
       ),
     };
-    if (field.backingType == field.typeName) {
+    if (!field.isCustomType) {
```
isoos, line 317: nit: maybe `isBuiltinType` would be more natural here

typed_sql/lib/src/codegen/build_code.dart
```diff
@@ -488,6 +488,63 @@ Iterable<Spec> buildTable(ParsedTable table, ParsedSchema schema) sync* {
...
+              > are nullable. Providing `null` will insert `NULL` for
+              > ${rowClass.fields.whereDefaultAndNullable.map((f) => '`${f.name}`').join(', ')}.
+              '''}
+
+              Returns a [InsertSingle] statement on which `.execute` must be
+              called for the row to be inserted.
+            ''')
+            ..optionalParameters.addAll(
+              rowClass.fields.map((field) {
+                // Field is optional if it has a default value xor is nullable,
```
isoos: Nit: Consider rephrasing this by putting the why it is required the main message / first line. (The xor logic doesn't need that much explanation).

typed_sql/lib/src/codegen/build_code.dart
```diff
@@ -488,6 +488,63 @@ Iterable<Spec> buildTable(ParsedTable table, ParsedSchema schema) sync* {
...
+                // but if it's both nullable and has a default value, then we
+                // will make it required! Because the default value of null will
+                // be read as NULL in the database and it won't get the database
+                // default value.
+                // NOTE: We do not set the default value as _default value_ in
+                //       dart, because this does not work for auto-increment or
+                //       for NOW() and similar annotations.
+                final isOptional = field.hasDefault ^ field.isNullable;
+                final nullable = field.hasDefault || field.isNullable
```
isoos: nit: `nullablePostfix` since `nullable` is also an input to it.

isoos (approved):

typed_sql/lib/src/codegen/build_code.dart
jonasfj: I decided to really enumerate all cases instead, this is complex, and future me will appreciate all the cases being enumerated.

PR #363, opened 2026-05-04
Upsert

jonasfj:
First commit is changes, the second commit is regenerated code.


typed_sql/lib/src/dialect/postgres_dialect.dart
```diff
@@ -25,7 +25,7 @@ SqlDialect postgresDialect() => _PostgresDialect();
 String _literal(Object? value) => switch (value) {
   null => 'NULL',
   true => 'TRUE',
-  false => 'TRUE',
+  false => 'FALSE',
```
isoos, line 28: yay 💯

typed_sql/lib/src/dialect/sqlite_dialect.dart
```diff
@@ -102,6 +102,40 @@ final class _Sqlite extends SqlDialect {
   @override
   (String, List<Object?>) insertInto(InsertStatement statement) {
     final resolver = ExpressionResolver(StatmentContext());
+    final alias = resolver.tableAlias;
+
+    String? conflictClause;
```
isoos, line 107: Nit: since this is roughly the same as the postgres version, wouldn't it be worthwhile to move it to `shared_dialect.dart`?

isoos (approved):

typed_sql/lib/src/dialect/sqlite_dialect.dart
jonasfj: some day, I don't want to try it yet.

If you want to refactor the dialects into using a shared mixin or something reasonable, you are invited, for now I think there is value in being dumb :D

It's not easy to spot what can be shared and what can't, ideally we could have a CommonSqlDialect that both postgres and sqlite subclass and then they only override a few specific methods.

But I'm not interested in doing that work :D

PR #364, opened 2026-05-05
Fix json path

jonasfj:
This change has some impact on what is actually possible, notably postgres will refuse some complex group-by queries if we have parameters in the expression here. So I figured we better not.

```diff
-    return 'JSON_EXTRACT(${expr(root)}, ${context.addParameter(path)})';
+    return 'JSON_EXTRACT(${expr(root)}, ${_escapeStringLiteral(path)})';
```

Granted group by something you've extracted from a JSON object is also a bit sketchy to begin with. But removing a limitation is always nice, and the way we've made the expression graph you cannot extract anything using a generic Expr, only using literal string or int (for list).

### DIFF
--- a/typed_sql/lib/src/dialect/mysql_dialect.dart
+++ b/typed_sql/lib/src/dialect/mysql_dialect.dart
@@ -29,15 +29,76 @@ String _literal(dynamic value) => switch (value) {
   false => 'FALSE',
   int i => i.toString(),
   double d => d.toString(),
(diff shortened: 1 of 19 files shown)


jonasfj:
Tests are mostly AI generated stuff, don't over think that part.

isoos (approved):
I see repeating pattern here that could be refactored 🙈

PR #366, opened 2026-05-08
Insert values mapped

jonasfj:


typed_sql/lib/typed_sql.dart
```diff
@@ -31,6 +31,7 @@ export 'src/typed_sql.dart'
         BinaryOperationExpression,
         CastExpression,
         ColumnDefinition,
+        ColumnType,
```
isoos, line 34: At one point it would be worth to separate these hidden types through files instead.

typed_sql/lib/src/typed_sql.statements.dart
```diff
@@ -49,20 +49,39 @@ final class SelectStatement extends SqlStatement {
...
   InsertStatement._(
     this.table,
-    this.columns,
     this.values,
     this.onConflict,
     this.returning,
   );
 }
 
+sealed class ValuesSource {
```
isoos, line 65: Let's add documentation to these.

jonasfj: I'm decidedly open to better class names, too.

But this is semi internal interface between us and Dialect.

typed_sql/lib/typed_sql.dart
jonasfj: yes and no, it's possible but there is a lot of stuff in there that has to be private. We could expose such properties using extension method that we don't export. But I don't think it's trivial.

isoos (approved):
While this is a bit overwhelming, I see no immediate bad thing in it. Let's try it out 🙈

PR #367, opened 2026-05-19
Refactor overrides

jonasfj:
I might still need some tests, do suggest reviewing commits individually.


typed_sql/lib/src/typed_sql.annotations.dart
```diff
@@ -262,24 +262,73 @@ final class Unique {
   const Unique.field({String? name}) : _name = name, _fields = null;
 }
 
+enum Nameing {
```
isoos: Rename this to `Naming`.

isoos (approved):

PR #372, opened 2026-05-21
More testing of nested transactions

jonasfj:
Covering https://github.com/dart-lang/pub-dev/pull/9398

### DIFF
--- a/typed_sql/test/typed_sql/transact/nested_basic/nested_basic_test.dart
+++ b/typed_sql/test/typed_sql/transact/nested_basic/nested_basic_test.dart
@@ -41,6 +41,36 @@ void main() {
     },
   );
 
+  r.addTest('successful nested transaction (savepoint)', (db) async {
+    await db.accounts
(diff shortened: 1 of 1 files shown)


typed_sql/test/typed_sql/transact/nested_basic/nested_basic_test.dart
```diff
@@ -41,6 +41,40 @@ void main() {
...
+      try {
+        await db.transact(() async {
+          // Update in inner transaction
+          await db.accounts
+              .byAccountNumber('0001')
+              .update((a, set) => set(balance: toExpr(300.0)))
+              .execute();
+        });
+      } on TransactionAbortedException {
+        // Expected
```
isoos: Nit: not clear if this is expected to happen (and we should test for it), or we should escalate?

jonasfj: probably a copy/paste bug, I should fix that 🙈

isoos (approved):

PR #374, opened 2026-06-03
Fix equality helpers + reorder foreign key equals expressions.

isoos:
- Switches the `ExpressionNullable<Type>` and `Expression<Type>` extensions' `.equals` and `.equalsValue` logic: non-nullable can get only the non-nullable value parameters.
- Updates the code generator's equals generator: we select the more lenient expression (if the FK is nullable, we use that, otherwise the referenced PK expression) as the lead object for the `.equals` call, that way if one of them is nullable but the other is not, we still generate type-safe and correct code.
- Updated tests.
- Verified pub-dev's use case, confirmed that it removes the COALESCE expressions without further modification.

### DIFF
--- a/typed_sql/lib/src/typed_sql.expr_ext.dart
+++ b/typed_sql/lib/src/typed_sql.expr_ext.dart
@@ -195,19 +195,19 @@ extension ExpressionNullableNum<T extends num> on Expr<T?> {
   ///  * [isNotDistinctFrom], to get `NULL` equivalent to `NULL`, or,
   ///  * [equalsUnlessNull], to explicitely get the SQL `=` semantics.
   /// {@endtemplate}
-  Expr<bool> equals(Expr<T> other) =>
+  Expr<bool> equals(Expr<T?> other) =>
       ExpressionEquals(this, other).orElseValue(false);
 
   /// {@macro equals}
-  Expr<bool> equalsValue(T other) => equals(toExpr(other));
+  Expr<bool> equalsValue(T? other) => equals(toExpr(other));
 
   /// {@template notEquals}
   /// Compare this expression to [other] using `<>` in SQL.
   /// {@endtemplate}
-  Expr<bool> notEquals(Expr<T> other) => equals(other).not();
+  Expr<bool> notEquals(Expr<T?> other) => equals(other).not();
 
   /// {@macro notEquals}
-  Expr<bool> notEqualsValue(T other) => notEquals(toExpr(other));
+  Expr<bool> notEqualsValue(T? other) => notEquals(toExpr(other));
 
   /// {@template isNotDistinctFrom}
   /// Compare this expression to [other] using `IS NOT DISTINCT FROM`.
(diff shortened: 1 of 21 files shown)


typed_sql/lib/src/typed_sql.expr_ext.dart
```diff
@@ -195,19 +195,19 @@ extension ExpressionNullableNum<T extends num> on Expr<T?> {
   ///  * [isNotDistinctFrom], to get `NULL` equivalent to `NULL`, or,
   ///  * [equalsUnlessNull], to explicitely get the SQL `=` semantics.
   /// {@endtemplate}
-  Expr<bool> equals(Expr<T> other) =>
+  Expr<bool> equals(Expr<T?> other) =>
```
jonasfj, line 198: why this?

I think we should instead remove this method in a follow up.

typed_sql/lib/src/typed_sql.expr_ext.dart
```diff
@@ -195,19 +195,19 @@ extension ExpressionNullableNum<T extends num> on Expr<T?> {
...
 
   /// {@macro equals}
-  Expr<bool> equalsValue(T other) => equals(toExpr(other));
+  Expr<bool> equalsValue(T? other) => equals(toExpr(other));
 
   /// {@template notEquals}
   /// Compare this expression to [other] using `<>` in SQL.
   /// {@endtemplate}
-  Expr<bool> notEquals(Expr<T> other) => equals(other).not();
+  Expr<bool> notEquals(Expr<T?> other) => equals(other).not();
```
jonasfj, line 207: probably we should also force this to use `<>` :)

jonasfj: okay, this is orthogonal, but something we should consider doing separately.

typed_sql/lib/src/typed_sql.expr_ext.dart
isoos: The nullable numeric expression should be matched with the same nullable expression. Agreed to remove in a follow-up, keeping this for now to keep scope in check.

jonasfj: well, now that I'm really considering, this does not and will not work.

Because `COALESCE(NULL = NULL, FALSE) IS NOT DISTINCT FROM FALSE` 

This `COALESCE(NULL = NULL, FALSE)` will eval to `FALSE`.

Which is correct SQL semantics, but different from what you'd expect in Dart, which is why there is an `.equalsUnlessNull` operator that will allow you to do this, if that is the behavior you want.

typed_sql/lib/src/typed_sql.expr_ext.dart
```diff
@@ -598,17 +598,17 @@ extension ExpressionNullableJsonValue on Expr<JsonValue?> {
 /// Extension methods for [bool] expressions.
 extension ExpressionBool on Expr<bool> {
   /// {@macro equals}
-  Expr<bool> equals(Expr<bool?> other) =>
-      ExpressionEquals(this, other).orElseValue(false);
+  Expr<bool> equals(Expr<bool> other) =>
+      ExpressionEquals(this, other).asNotNull();
```
jonasfj, line 602: This is safe, let's keep it, right?

typed_sql/lib/src/typed_sql.expr_ext.dart
```diff
@@ -289,17 +289,17 @@ extension ExpressionNullableBool on Expr<bool?> {
   Expr<bool> orElseValue(bool value) => orElse(toExpr(value));
 
   /// {@macro equals}
-  Expr<bool> equals(Expr<bool> other) =>
+  Expr<bool> equals(Expr<bool?> other) =>
       ExpressionEquals(this, other).orElseValue(false);
```
jonasfj, line 293: this is unsafe, do we need it?

isoos:
Closing in favor of #381

PR #378, opened 2026-06-10
Typed sql literal migration

jonasfj:


typed_sql/lib/src/adapter/postgres_adapter.dart
```diff
@@ -475,8 +475,8 @@ List<Object?> _paramsForPostgres(List<Object?> params) => params
         // Negative years and year zero most be written as BC in postgres
         // https://www.postgresql.org/docs/17/datetime-input-rules.html
         DateTime d when d.year <= 0 =>
-          '${d.copyWith(year: 1 - d.year).toIso8601String()} BC',
-        DateTime d => d.toIso8601String(),
+          '${d.toUtc().copyWith(year: 1 - d.toUtc().year).toIso8601String()} BC',
```
isoos, line 478: Is this literal string or parameter? I would hope parameters can be sent in as-is.

typed_sql/lib/src/typed_sql.expr.dart
```diff
@@ -454,95 +476,199 @@ final class CastExpression<T, R> extends SingleValueExpr<R> {
...
+              'value',
+              'Only String, int, double, bool, null, DateTime, Uint8List,'
+                  ' JsonValue values are allowed',
+            );
+        }
+    }
+  }
+}
+
+final class ValueExpression<T> extends SingleValueExpr<T> {
```
isoos, line 583: I'm wondering: `ValueExpression` extending `SingleValueExpr` seems a bit confusing, both being Value and Expr[ession]... maybe `ParameterExpression` would be a better name?

typed_sql/lib/src/adapter/postgres_adapter.dart
jonasfj: This is when sending parameters... yes, we sadly encode `DateTime` as string.

probably we should try `TypedValue` :see_no_evil: 

But it's an orthogonal change.

typed_sql/lib/src/typed_sql.expr.dart
jonasfj: Yes, `ParameterExpression` would be a better name.

I think we should consider doing such rename later. Because when/if we want to do _prepared statements_ you'll want some sort of Parameter thing too that isn't bound to a value, which is what these parameters are.

But also: this is internal naming only relevant to dialect implementers. Actual users of `typed_sql` will never hold a `ValueExpression`, it'll always be given to them as `Expr<T>` of some `T`.

isoos: Ack, good to do it later.

isoos (approved):

PR #379, opened 2026-06-11
Allow nullable boolean in JOIN conditions.

isoos:


typed_sql/test/typed_sql/join/nullable_on/nullable_on_test.dart
```diff
@@ -0,0 +1,102 @@
...
+      final c = x.$1.compareTo(y.$1);
+      return c != 0 ? c : x.$2.compareTo(y.$2);
+    });
+  }
+
+  for (final c in _cases) {
+    r.addTest(c.name, (db) async {
+      check(await joinOn(db, c.on)).deepEquals(c.expected);
+    });
+  }
```
jonasfj:
```suggestion
  for (final c in _cases) {
    r.addTest(c.name, (db) async {
      var pairs = await db.items
          .join(db.items)
          .on(c.on)
          .select((a, b) => (a.id, b.id))
          .fetch();
      pairs = pairs.toList()..sort((x, y) {
        final c = x.$1.compareTo(y.$1);
        return c != 0 ? c : x.$2.compareTo(y.$2);
      });
      check(pairs).deepEquals(c.expected);
    });
  }
```

Do also consider if we can just use orderby, instead of this complex sort logic...

```suggestion
  for (final c in _cases) {
    r.addTest(c.name, (db) async {
      final pairs = await db.items
          .join(db.items)
          .on(c.on)
          .select((a, b) => (a.id, b.id))
          .orderBy((a, b) => [
            (a, .ascending),
            (b, .ascending),
          ])
          .fetch();
      check(pairs).deepEquals(c.expected);
    });
  }
```

jonasfj (approved):
Please also update doc/ where appropriate

typed_sql/test/typed_sql/join/nullable_on/nullable_on_test.dart
isoos: Done.

jonasfj (approved):

PR #381, opened 2026-06-11
Updated equals() code generation + removals from nullable expressions.

isoos:


isoos:
Note: I've also verified this on pub-dev, and with the other changes, this now removes the `COALESCE` expressions from our queries.

jonasfj:
I think we landed this in https://github.com/google/dart-neats/pull/382

PR #383, opened 2026-06-21
sanitize_html: allow <picture> and <source> elements

chaudharydeepanshu:
Towards dart-lang/pub-dev#5923.

@jonasfj's condition there was to only add tags backed by a referenceable allowlist, not ad-hoc ones. That exists now: `gjtorikian/html-pipeline`'s allowlist (what this package's doc links to) lists `picture`, `source`, and `source → srcset`. This mirrors it — `srcset` is validated like `img`'s `src`, so a bad URL in the set drops just the attribute.

Also refreshes the doc link to html-pipeline's current path (`jch/` → `gjtorikian/`).


sanitize_html/lib/src/sane_html_validator.dart
```diff
@@ -177,6 +179,17 @@ bool _validUrl(String url) {
   }
 }
 
+// `srcset` is a comma-separated list of "<url> [descriptor]" candidates.
+// Validate every candidate's URL with the same scheme rules as `src`, so a
+// single bad entry (e.g. a `javascript:` URL) rejects the whole attribute.
```
jonasfj:
```suggestion
// `srcset` is a comma-separated list of "<url> [descriptor]" candidates.
// Validate every candidate's URL with the same scheme rules as `src`, so a
// single bad entry (e.g. a `javascript:` URL) rejects the whole attribute.
//
// See also: https://html.spec.whatwg.org/multipage/images.html#srcset-attributes
```

sanitize_html/lib/src/sane_html_validator.dart
```diff
@@ -177,6 +179,17 @@ bool _validUrl(String url) {
...
+bool _validSrcset(String value) {
+  for (final candidate in value.split(',')) {
+    final url = candidate.trim().split(RegExp(r'\s+')).first;
```
jonasfj: Use a top-level regex if you must, but maybe also look at html5 parsing rules.

I don't think we should allow arbitrary descriptions:
https://html.spec.whatwg.org/multipage/images.html#srcset-attributes

jonasfj:
My reservation has mainly been that I don't want to maintain an arbitrary set of rules.
And that I don't want a lot of PRs asking to allow this and that.

I think that in general people are probably better off making a fork.
And that we keep this package simple and manageable. Because it's hard to a sound opinion on every HTML tag and attribute :D

---------------

That said, I'll admit that probably we also have to be practical once in a while, and that stuff that we think is important. Even if it means deviating from `gjtorikian/html-pipeline`.

In any case, I'm super happy that this just bring us into line with `html-pipeline`.

jonasfj (commented):
Please consider if we should be more strict in the `srcset` attribute validation.

And run `dart format`.

sanitize_html/lib/src/sane_html_validator.dart
chaudharydeepanshu: Took the spec link; extended the comment a line for the descriptor rule, so left the suggestion unapplied.

sanitize_html/lib/src/sane_html_validator.dart
chaudharydeepanshu: Done. A top-level `_srcsetDescriptor` regex now requires a width (`640w`) or density (`1.5x`); anything else drops the attribute. Kept the comma split rather than the full WHATWG parser since `_validUrl` rejects `data:`, so candidate URLs never contain commas here. Added tests for the descriptor cases.

chaudharydeepanshu:
Pushed an update: stricter descriptor validation + dart format. Details inline.

jonasfj (approved):

Issue #250, opened 2024-09-02
Dartdoc_test reports no tests run when everything works


sigurdm:
* Explain the issue,
 * Using bullet points!

```dart
/// This class is excellent
/// ```dart
/// void main() {
///   print('hi');
///   main();
/// }
/// ```
class Foo {}
```

```
> dart test
00:01 +0: loading test/dartdoc_test.dart                                                                                       Extracting code samples ...
Analyzing code samples ...
No tests ran.                                                                                                                   
No tests were found.
```

I would have expected at least a single test to have run. If I make a syntax error it does indeed report a failed test.

jonasfj:
@takumma FYI, in case you're interested.

Hira-Iftikhar-123:
The issue I analyzed is that the dartdoc_test doesn't recognize or detect the test running inside the documentation.

Solution I proposed:

Modify test/dartdoc_test.dart to explicitly scan and run the code samples in lib/abd.dart

syntax:

import 'package:dartdoc_test/dartdoc_test.dart';

// main definition
void main ()
{
    testDartdocExamples('lib/abd.dart');
}

Why this is helpful:

The tool might not automatically detect the test, so explicitly registering it forces dartdoc_test to analyze lib/abd.dart.

to run:

dart test

PR #393, opened 2026-07-13
fix: use p.join for Windows path compatibility (#250)

assassinaj602:
## Description
Fixes the "Path must be normalized" error on Windows when running `dartdoc_test`.

## The Problem
The test directory path was hardcoded with forward slashes in `lib/src/resource.dart`:
```dart
const _testPath = '.dart_tool/dartdoc_test';  // ❌ Causes mixed slashes on Windows


jonasfj:
dartdoc_test is discontinued.

PR #404, opened 2026-07-30
Support more index types + covering columns.

isoos:


jonasfj:
What is the motivation here?

I strongly dislike database specific features.

isoos:
The main feature is to support GIN and GIST indexes (for JSON index and for full-text search), and while we are at it,  covering columns as well (which is probably a less used feature, but still a useful one).

